PR #15, opened 2024-09-30
Eval utilities

peadaroh:
Superseded by:
https://github.com/humanloop/humanloop-python/pull/18


andreibratu (dismissed):
LGTM - I'd prefer changing the '-Params' prefix on classes to the full '-Parameters'

jordn:
10 minutes of loom feedback just on the PR description!

https://www.loom.com/share/fd1d783081c14ae8829171fc2022f2e9

One, this feature's a huge unlock, and looks great. Things I'm arguing over:

1. `pipeline` is confusing to me. New term which won't work that well for other files. 
2.   Lets put `function` or `callable` into the `file` object? That looks consistent with the nice DX for having a external evaluator you have here
3. Some nits about where to put the link and how to lay out the table -- it's unclear to me which is is control, which is latest for example.

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        if local_evaluators["args_type"] == "target_required":
+            requires_target = True
+            break
+    if requires_target:
+        missing_target = 0
+        for datapoint in dataset.datapoints:
+            if not datapoint.target:
+                missing_target +=1
+        if missing_target > 0:
+            raise ValueError(f"Datapoint {datapoint.id} has no target is required for this Evaluator: {local_evaluator['path']}")
```
andreibratu: Why not raise directly in the body of the for? datapoint will point to the latest point in the dataset, not the one missing the `target`

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+    # Get or create the Evaluation based on the name
+    evaluation = None
+    try:
+        evaluation = client.evaluations.create(
+            name=name,
+            dataset={"file_id": hl_dataset.id},
+            evaluators=[{"path": e["path"]} for e in evaluators],
+            file={"id": file.id},
+        )
+    except ApiError as error_:
```
andreibratu, line 252: why the undercase? I don't see any other variable `error` clashing with this one

PeadarOhAodha:
Haven't watched the loom yet, but:

> pipeline is confusing to me. New term which won't work that well for other files.

Agree, i had it as callable on Friday, changed to pipeline based on some comments from Raza in demo. Prefer `callable` of `function`. Wil lchange

> Lets put function or callable into the file object? That looks consistent with the nice DX for having a external evaluator you have here`

yes also agree

> Some nits about where to put the link and how to lay out the table -- it's unclear to me which is is control, which is latest for example.

will improve this

src/humanloop/eval_utils.py
peadaroh: Good spot. The message in the ValueError is just wrong - it should talk about the number of missings. I changed the logic and didn't change the error.

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        case "flow":
+            # Be more lenient with Flow versions as they are arbitrary json
+            try:
+                FlowKernel.parse_obj(version)
+            except ValidationError:
+                version = {"attributes": version}
+                file_dict = {**file, **version}
+            file = client.flows.upsert(**file_dict)
+        case "prompt":
+            file = client.flows.upsert(**file_dict)
```
andreibratu: why call `.flows` when type is "prompt" or "tool"?

src/humanloop/eval_utils.py
peadaroh: error is a reserved keyword in python

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+            raise NotImplementedError(f"Unsupported File version: {type_}")
+
+
+def get_score_from_evaluator_stat(stat: NumericStats | BooleanStats) -> float | None:
+    """Get the score from an Evaluator Stat."""
+    score = None
+    match stat:
+        case BooleanStats():
+            if stat.total_logs:
+                score = round(stat.num_true / stat.total_logs, 2)
```
andreibratu: Is it possible for a dataset to have no logs and add it to the evaluation? Wondering if this could result in a divide by zero

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        case NumericStats():
+            score = round(stat.mean, 2)
+        case _:
+            raise ValueError("Invalid Evaluator Stat type.")
+    return score
+
+
+def _progress_bar(total: int, progress: int):
```
andreibratu, line 440: The usecase is simple indeed, but I think it's worth considering switching to tqdm so we don't have to support this helper too

andreibratu: https://github.com/tqdm/tqdm

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        _progress_bar.start_time = None
+
+    sys.stdout.write(progress_display)
+    sys.stdout.flush()
+
+    if progress >= total:
+        sys.stdout.write("\n")
+
+
+def get_evaluator_stats_by_path(
```
andreibratu, line 473: public function after private `_progress_bar`

src/humanloop/eval_utils.py
peadaroh: i intetionally didn't use tqdm as it would mean adding additional packages to the SDK, which i want to avoid

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+    """The change in score since the previous version of your pipeline for a specific Evaluator."""
+    threshold: float | None
+    """The threshold to check the Evaluator against."""
+    threshold_check: bool | None
+    """Whether the latest version has an average Evaluator result above a threshold."""
+
+# TODO: Explore restarting runs, but without having to regenerate logs
+
+
+def _run_eval(
```
andreibratu, line 153: _run_eval is private and not used in this module

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        else:
+            print(
+                f"{RED}❌ Latest score [{score}] below the threshold [{threshold}] for evaluator {evaluator_path}.{RESET}"
+            )
+            return False
+    else:
+        raise ValueError(f"Evaluator {evaluator_path} not found in the stats.")
+
+
+def check_evaluation_improvement(
```
andreibratu, line 518: The two `check_` functions are public, but seem to be used only in the context of this module (by a private function)

src/humanloop/eval_utils.py
peadaroh: its a minor abuse of notation; but we want users to only use this via the client evaluations.run (see client.py). So indicating it is private here i think is helpful in that regard

src/humanloop/eval_utils.py
peadaroh: Users may want to import these and customise them

src/humanloop/eval_utils.py
andreibratu: Per irl discussion, add a comment explaining the use of the function and close this thread

src/humanloop/eval_utils.py
andreibratu: ack closing

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        client=client,
+        type_=type_,
+        file_id=file.id,
+        version_id=file.version_id,
+        evaluation_id=evaluation.id,
+        batch_id=batch_id
+    )
+
+    # Define the function to execute your pipeline in parallel and Log to Humanloop
+    def process_datapoint(datapoint):
```
andreibratu: nittest nit: add type hinting `DatapointDict`

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+        #  Why are both `id` and `version_id` needed in the API?
+        "id": file_id,
+        "version_id": version_id,
+        "evaluation_id": evaluation_id,
+        "batch_id": batch_id,
+    }
+    match type_:
+        case "flow":
+            return partial(client.flows.log, **log_request, trace_status="complete")
+        case "prompt":
```
andreibratu, line 399: case "prompt" | "evaluator" | "tool":

src/humanloop/eval_utils.py
andreibratu: ack closing

src/humanloop/eval_utils.py
andreibratu: resolving per discussion on _run_eval

src/humanloop/eval_utils.py
andreibratu: closing per discussion on _run_eval

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+from .requests import CodeEvaluatorRequestParams as CodeEvaluatorDict
+from .requests import LlmEvaluatorRequestParams as LLMEvaluatorDict
+from .requests import HumanEvaluatorRequestParams as HumanEvaluatorDict
+
+
+# Responses are Pydantic models
+from .types import FlowKernelRequest as FlowKernel
+from .types import BooleanEvaluatorStatsResponse as BooleanStats
+from .types import NumericEvaluatorStatsResponse as NumericStats
+from .types import UpdateDatesetAction as UpdateDatasetAction  # TODO: fix original type typo
```
andreibratu, line 41: Is this todo still relevant?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,543 @@
...
+    evaluator_stats_by_path = get_evaluator_stats_by_path(
+        stat=stats.version_stats[-1],
+        evaluation=evaluation
+    )
+    if evaluator_path in evaluator_stats_by_path:
+        evaluator_stat = evaluator_stats_by_path[evaluator_path]
+        score = get_score_from_evaluator_stat(stat=evaluator_stat)
+        if score >= threshold:
+            print(
+                f"{GREEN}✅ Latest eval [{score}] above threshold [{threshold}] for evaluator {evaluator_path}.{RESET}"
```
andreibratu, line 506: can we make the text consistent with the below threshold print? e.g. latest eval vs latest score

src/humanloop/requests/evaluated_version_response.py
```diff
@@ -2,10 +2,10 @@
...
 import typing
 from .prompt_response import PromptResponseParams
-from ..types.tool_response import ToolResponse
+from .tool_response import ToolResponseParams
 from .evaluator_response import EvaluatorResponseParams
 from .flow_response import FlowResponseParams
 
 EvaluatedVersionResponseParams = typing.Union[
-    PromptResponseParams, ToolResponse, EvaluatorResponseParams, FlowResponseParams
+    PromptResponseParams, ToolResponseParams, EvaluatorResponseParams, FlowResponseParams
```
andreibratu: Can we change "-Params" suffix to "-Parameters" on all responses?

andreibratu (changes requested):
* Fix the two small bugs with type and throwing on no  target datapoint
* Small pythonic nits
* MAybe change from -Params to -Parameters? No strong preference

src/humanloop/eval_utils.py
peadaroh: Fixed

src/humanloop/eval_utils.py
peadaroh: Fixed

src/humanloop/eval_utils.py
peadaroh: No that should never happen here

src/humanloop/eval_utils.py
peadaroh: Solved - but it was the pydantic response model not a dict

src/humanloop/eval_utils.py
peadaroh: Yes - see the typo:
UpdateDatesetAction

src/humanloop/eval_utils.py
peadaroh: This will soon be replaced by our server side version so don't want to change further atm

src/humanloop/requests/evaluated_version_response.py
peadaroh: This is autogenerated from the SDK

src/humanloop/base_client.py
```diff
@@ -0,0 +1,178 @@
...
+from .evaluators.client import AsyncEvaluatorsClient
+from .flows.client import AsyncFlowsClient
+from .files.client import AsyncFilesClient
+from .evaluations.client import AsyncEvaluationsClient
+from .logs.client import AsyncLogsClient
+
+
+class BaseHumanloop:
+    """
+    Use this class to access the different functions within the SDK. You can instantiate any number of clients with different configuration that will propagate to these functions.
```
harry-humanloop, line 30: where is this taken from?

(would expect some comment about it in the readme or something)

src/humanloop/base_client.py
```diff
@@ -0,0 +1,178 @@
...
+        self.files = AsyncFilesClient(client_wrapper=self._client_wrapper)
+        self.evaluations = AsyncEvaluationsClient(client_wrapper=self._client_wrapper)
+        self.logs = AsyncLogsClient(client_wrapper=self._client_wrapper)
+
+
+def _get_base_url(*, base_url: typing.Optional[str] = None, environment: HumanloopEnvironment) -> str:
+    if base_url is not None:
+        return base_url
+    elif environment is not None:
+        return environment.value
```
harry-humanloop, line 176: this doesn't look like it can happen

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+
+# ANSI escape codes for colors
+YELLOW = "\033[93m"
+CYAN = "\033[96m"
+GREEN = "\033[92m"
+RED = "\033[91m"
+RESET = "\033[0m"
+
+
+class Identifiers(TypedDict, total=False):
```
harry-humanloop: `total=False` makes keys not required by default; i think the keys below don't need the NotRequired then? any reason for this verbosity?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    id: NotRequired[str]
+    """The ID of the File on Humanloop."""
+    path: NotRequired[str]
+    """The path of the File on Humanloop."""
+
+
+class File(Identifiers, total=False):
+    """A File on Humanloop (Flow, Prompt, Tool, Evaluator)."""
+    type: NotRequired[Literal["flow", "prompt", "tool", "evaluator"]]
+    """The type of File this pipeline relates to on Humanloop."""
```
harry-humanloop: pipeline - do we want to keep using this word?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    """Common identifiers for the objects required to run an Evaluation."""
+    id: NotRequired[str]
+    """The ID of the File on Humanloop."""
+    path: NotRequired[str]
+    """The path of the File on Humanloop."""
+
+
+class File(Identifiers, total=False):
+    """A File on Humanloop (Flow, Prompt, Tool, Evaluator)."""
+    type: NotRequired[Literal["flow", "prompt", "tool", "evaluator"]]
```
harry-humanloop: can we use an enum from the sdk here instead?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    """The type of File this pipeline relates to on Humanloop."""
+    version: NotRequired[Version]
+    """The contents uniquely define the version of the File on Humanloop"""
+    function: Callable
+    """The function being evaluated.
+    It will be called using your Dataset `inputs` as follows: `output = pipeline(**datapoint.inputs)`.
+    If `messages` are defined in your Dataset, then `output = pipeline(**datapoint.inputs, messages=datapoint.messages)`.
+    It should return a single string output. If not, you must provide a `custom_logger`.
+    """
+    custom_logger: NotRequired[Callable]
```
harry-humanloop, line 94: could be a Protocol perhaps? (to elaborate on the arguments)

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    """function that logs the output of your pipeline to Humanloop
+    If provided, it will be called as follows:
+        ```
+        output = pipeline(**datapoint.inputs).
+        log = custom_logger(client, output)
+        ```
+        Inside the custom_logger, you can use the Humanloop `client` to log the output of your pipeline.
+        If not provided your pipline must return a single string.
```
harry-humanloop, line 102: hmm, looking at this docstring, it looks like this custom_logger replaces the logger, and isn't in-addition-to the default logger.
should call it out that this would be used in place of a default humanloop logger

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    """optional function that logs the output judgment from your Evaluator to Humanloop, if provided, it will be called as follows:
+    ```
+    judgment = callable(log_dict)
+    log = custom_logger(client, judgmemt)
+    ```
+    Inside the custom_logger, you can use the Humanloop `client` to log the judgment to Humanloop.
+    If not provided your pipline must return a single string.
+    """
+    threshold: NotRequired[float]
+    """The threshold to check the Evaluator against."""
```
harry-humanloop: should i be able to tell from this docstring what the behaviour is when this is set and the threshold isn't met? also is this a lower bound or upper bound?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    try:
+        function_ = file.pop("function")
+    except KeyError as _:
+        raise ValueError("You must provide a function to run your Evaluation.")
+    try:
+        type_ = file.pop("type")
+    except KeyError as _:
+        # Default to flows if not type specified
+        type_ = "flow"
+        logger.warning("No type specified, defaulting to 'flow'.")
```
harry-humanloop, line 185: why is this a warning?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+                    end_time=datetime.now(),
+                )
+                logger.warning(f"\nEvaluator {local_evaluator['path']} failed with error {str(e)}")
+
+    # Execute the pipeline and send the logs to Humanloop in parallel
+    total_datapoints = len(hl_dataset.datapoints)
+    print(f"\n{CYAN}Navigate to your Evals:{RESET} {evaluation.url}")
+    print(f"{CYAN}Version Id: {file.version_id}{RESET}")
+    print(f"{CYAN}Run Id: {batch_id}{RESET}")
+    print(f"{CYAN}\nRunning your pipeline over the Dataset...{RESET}")
```
harry-humanloop: i suspect there are best practices that we should follow here - like omitting the color code for terminals that don't support them etc. or letting people control where this is printed to.

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+        for _ in as_completed(futures):
+            completed_tasks += 1
+            _progress_bar(total_datapoints, completed_tasks)
+
+    # Wait for the Evaluation to complete then print the results
+    complete = False
+    stats = None
+    while not complete:
+        stats = client.evaluations.get_stats(id=evaluation.id)
+        sys.stdout.write(stats.progress)
```
harry-humanloop: why sys.stdout.write and not print?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+                delta=delta,
+                threshold=threshold,
+                threshold_check=threshold_check,
+            )
+        )
+    return checks
+
+
+def _get_log_func(
+        client: BaseHumanloop,
```
harry-humanloop: whats with the indentation here

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+        type_: Literal["flow", "prompt", "tool", "evaluator"],
```
harry-humanloop: this type (flow | prompt | tool | evaluator) should be defined in a single place (if not available in SDK)

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    progress_display = f"\r[{bar}] {progress}/{total}"
+    progress_display += f" ({percentage:.2f}%)"
+
+    if progress < total:
+        progress_display += f" | ETA: {int(eta)}s"
+    else:
+        progress_display += " | DONE"
+        _progress_bar.start_time = None
+
+    sys.stdout.write(progress_display)
```
harry-humanloop: this should probably go to stderr - things like logging should go to stderr; stdout is for the program's main output.
(tqdm defaults to stderr for its `file` argument that it writes to)

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+def check_evaluation_threshold(
+    evaluation: EvaluationResponse,
+    stats: EvaluationStats,
+    evaluator_path: str,
+    threshold: float,
+) -> bool:
+    """Checks if the latest version has an average Evaluator result above a threshold."""
+    # TODO: Update the API so this is not necessary
+    evaluator_stats_by_path = get_evaluator_stats_by_path(
+        stat=stats.version_stats[-1],
```
harry-humanloop: why [-1]? does that mean you can't kick of multiple of these at the same time?

src/humanloop/client.py
```diff
@@ -1,178 +1,50 @@
...
         )
-        self.prompts = PromptsClient(client_wrapper=self._client_wrapper)
-        self.tools = ToolsClient(client_wrapper=self._client_wrapper)
-        self.datasets = DatasetsClient(client_wrapper=self._client_wrapper)
-        self.evaluators = EvaluatorsClient(client_wrapper=self._client_wrapper)
-        self.flows = FlowsClient(client_wrapper=self._client_wrapper)
-        self.files = FilesClient(client_wrapper=self._client_wrapper)
-        self.evaluations = EvaluationsClient(client_wrapper=self._client_wrapper)
-        self.logs = LogsClient(client_wrapper=self._client_wrapper)
+        self.evaluations.run = partial(_run_eval, client=self)
```
harry-humanloop: i don't think this should be exposed like this, as a single long-running method among API requests.

also this conflicts with the runs concept we'll be introducing.

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+        ```
+        output = pipeline(**datapoint.inputs).
+        log = custom_logger(client, output)
+        ```
+        Inside the custom_logger, you can use the Humanloop `client` to log the output of your pipeline.
+        If not provided your pipline must return a single string.
+    """
+
+
+class RunDataset(Identifiers):
```
harry-humanloop, line 106: how does the `total=False` on `Identifiers` affect `RunDataset`?

is `datapoints` here `Required` or `NotRequired` by default?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    datapoints: Sequence[DatapointDict]
+    """The datapoints to map your pipeline over to produce the outputs required by the evaluation."""
+    action: NotRequired[UpdateDatasetAction]
+    """How to update the Dataset given the provided Datapoints; 
+    `set` replaces the existing Datapoints and `add` appends to the existing Datapoints."""
+
+
+class RunEvaluator(Identifiers):
+    """A to provide judgments for this Evaluation."""
```
harry-humanloop: A to

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    threshold: NotRequired[float]
+    """The threshold to check the Evaluator against."""
+
+
+class EvaluatorCheck(BaseModel):
+    """Summary data for an Evaluator check."""
+    path: str
+    """The path of the Evaluator used in the check."""
+    improvement_check: bool
+    """Whether the latest version of your pipeline has improved across for a specific Evaluator."""
```
harry-humanloop: improved across for a specific Evaluator?

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    if evaluators:
+        for evaluator in evaluators:
+            # If a callable is provided for an Evaluator, we treat it as External
+            eval_function = evaluator.get("function")
+            if eval_function is not None:
+                local_evaluators.append(evaluator)
+                spec = ExternalEvaluator(
+                    arguments_type=evaluator["args_type"],
+                    return_type=evaluator["return_type"],
+                    attributes={"code": inspect.getsource(eval_function)},
```
harry-humanloop, line 220: should call out that this is automatically done - risk of leaking creds.

src/humanloop/eval_utils.py
andreibratu: https://discuss.python.org/t/should-total-false-mean-all-keys-are-notrequired-by-default/37538

Yep, they are equivalent

andreibratu (commented):
Ty for addressing comments, have no further additions to harry's insights

andreibratu (approved):

src/humanloop/base_client.py
peadaroh: as linked to in PR comment and slack message:
https://buildwithfern.com/learn/sdks/features/augment-with-custom-code

src/humanloop/base_client.py
peadaroh: This is ferns auto generated client, so please mention to them

src/humanloop/eval_utils.py
peadaroh: Because we are making a choice when the user did not specify a type

src/humanloop/eval_utils.py
peadaroh: Have a TODO around using logger instead of print, obvious improvement

src/humanloop/eval_utils.py
peadaroh: I'm not sure yet if this is the pattern we want to follow here

src/humanloop/eval_utils.py
peadaroh: typo

src/humanloop/eval_utils.py
peadaroh: typo

src/humanloop/eval_utils.py
peadaroh: i'll tweak the type dicts

src/humanloop/client.py
peadaroh: > i don't think this should be exposed like this, as a single long-running method among API requests.

I think its ok; i like the cleanliness of accessing it like this as per our discussions on friday about the diff options.

Don't think it should clash with runs concept - you should elaborate

src/humanloop/eval_utils.py
peadaroh: the updated stats endpoint as part of your work should solve this as it will be pushed into the endpoint

src/humanloop/eval_utils.py
peadaroh: i can elaborate on the behaviour. Bounding isn't relevant

src/humanloop/eval_utils.py
peadaroh: we don't know what the type of the arg would be i suppose; depends on the function

src/humanloop/base_client.py
harry-humanloop: by "this" i meant the docstring here. the whitespace is weird (and there's seemingly a stray `from .environment import HumanloopEnvironment`).
i'm guessing this was copied from somewhere and i think we should add something to the readme (perhaps under a "contributing" section?) about how this was initialized

src/humanloop/eval_utils.py
harry-humanloop: yes from other docstrings, i can tell that this acts as a lower bound for acceptable values. (i probably was unclear about what i meant by bounding above).
think we have to mention the direction of the check (i.e. is being higher/lower than this value a pass or fail).

src/humanloop/eval_utils.py
harry-humanloop: hmm, i think it's more info-level but can see the argument for warning (to push people to make this explicit, but to not require knowledge about "flow" up front)

src/humanloop/client.py
harry-humanloop: the runs concept will probably be exposed as `humanloop.evaluations.create_run()`. having this `run()` method that does something extremely different would be confusing.

andreibratu: @harry-humanloop, can I get some pointers to the 'runs' concept? I'd like to get up to speed with it

src/humanloop/requests/tool_response.py
```diff
@@ -0,0 +1,131 @@
+# This file was auto-generated by Fern from our API Definition.
+
+from __future__ import annotations
+import typing_extensions
```
gavinfd-hl: Hmm a few duplicate imports here. I know this is auto-generated, just calling it out as weird.

src/humanloop/eval_utils.py
peadaroh: I'll define it separately within this file, as we don't have such an eum

src/humanloop/eval_utils.py
```diff
@@ -0,0 +1,555 @@
...
+    bar_length = 40
+    block = int(round(bar_length * progress / total))
+    bar = "#" * block + '-' * (bar_length - block)
+
+    percentage = (progress / total) * 100
+    elapsed_time = time.time() - _progress_bar.start_time
+    time_per_item = elapsed_time / progress if progress > 0 else 0
+    eta = (total - progress) * time_per_item
+
+    progress_display = f"\r[{bar}] {progress}/{total}"
```
harry-humanloop, line 458: @PeadarOhAodha i think this is the load-bearing carriage return that moves the cursor to the beginning of the current line, thus overwriting the line.

https://www.geeksforgeeks.org/difference-between-newline-and-carriage-return-in-python/
https://gist.github.com/alex-calderwood/f958ee4c5c0a96babef64673a8f1b323

peadaroh: thanks

PR #21, opened 2024-10-15
Prompt utils

harry-humanloop:
- Add `humanloop.prompts.populate_template`

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "humanloop"
-version = "0.8.6b3"
+version = "0.8.6"
 description = ""
 readme = "README.md"
 authors = []
--- a/src/humanloop/eval_utils.py
+++ b/src/humanloop/eval_utils.py
@@ -7,6 +7,7 @@ where you are managing the runtime of your application in your code.
 Functions in this module should be accessed via the Humanloop client. They should
 not be called directly.
 """
+
 import logging
 from datetime import datetime
 from functools import partial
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "humanloop"
-version = "0.8.6b3"
```
peadaroh, line 3: @harry-humanloop was this intentional?

src/humanloop/eval_utils.py
```diff
@@ -208,29 +214,29 @@ def _run_eval(
...
-        
+
     elif type_ == "prompt":
         try:
             _ = Prompt.parse_obj(version)
         except ValidationError as error_:
             logger.error(msg=f"Invalid Prompt `version` in your `file` request. \n\nValidation error: \n)")
             raise error_
         hl_file = client.prompts.upsert(**file_dict)
-
```
peadaroh, line 219: Are these ruff changes?

pyproject.toml
harry-humanloop: not really; this was from the first commit in this PR, by fern-bot, which i suspect comes from `fern generate`

(https://github.com/humanloop/humanloop-python/pull/21/commits/77b76f7dcd91cec808fdb6b7c4ae59cc3eac7799)

src/humanloop/eval_utils.py
harry-humanloop: all are ruff, apart from https://github.com/humanloop/humanloop-python/pull/21/commits/deefc8b45d89b9b0de7ef41a9659881d460770fc

PR #27, opened 2024-11-06
Eval runs

harry-humanloop:
Update eval utils to create run and pass run_id in log call.

Ignore the first commit by fernbot when reviewing


src/humanloop/requests/version_specification.py
```diff
@@ -0,0 +1,37 @@
+# This file was auto-generated by Fern from our API Definition.
+
+import typing_extensions
+import typing_extensions
```
peadaroh, line 4: Check where this is used

peadaroh (approved):
LGTM - just check the use of that legacy stuff

src/humanloop/requests/version_specification.py
harry-humanloop: no longer in SDK now with updated openapi spec

harry-humanloop:
pushed (accidentally) directly to master:
https://github.com/humanloop/humanloop-python/commit/46c6504a706d0bba1097c888cb56bfc96a7df556

PR #42, opened 2025-01-06
Flow complete fix

andreibratu:
* Simplified the code that kept track of the trace built by decorators - instead of relying on the global TRACE_CONTEXT variable, we are only peeking at the status of the ancestor on the level above
* Aligned the Python SDK with TS SDK: we now properly support spans created by LLM library streaming calls. To do this, the instrument spans have a complete bool property attached to them. The bool is false on span creation and turns true when the streaming ends. The Prompt span is no longer exported immediately after the decorated function exit, but only after all instrumentor spans underneath it are uploaded
* The exporter correctly completes flow traces - instead of waiting for the program exit to upload the spans, I am attaching a new span attribute instructing the exporter which spans' logs should be uploaded before completing the flow log. To make this work, the processor keeps a running list of spans as they start, which belong in a particular trace. The list is attached to the flow span during the onEnd signal. The spans created by utilities will no longer have a UUIDv4 name but be named after the log type they carry, e.g., `humanloop.prompt` or `humanloop.flow`. This is also more dogmatic with how OT does things
* Renamed the decorators module to utilities to align with the name used in TS SDK

Things to look out for: I am introducing some async wait logic for the streaming case in the processor. I use a thread pool to not block the main thread. Please double-check if it makes sense and if there are no gotchas


tests/utilities/test_flow_decorator.py
```diff
@@ -135,18 +137,7 @@ def test_decorators_with_flow_decorator(
     # THEN the span are returned bottom to top
     assert read_from_opentelemetry_span(span=spans[1], key=HUMANLOOP_FILE_KEY)["tool"]
     assert read_from_opentelemetry_span(span=spans[2], key=HUMANLOOP_FILE_KEY)["prompt"]
-    # assert read_from_opentelemetry_span(span=spans[3], key=HL_FILE_OT_KEY)["flow"]
```
andreibratu, line 138: No longer relying on TRACE_FLOW_CONTEXT here, using the ancestor to determine that logs are uploaded correctly.

tests/utilities/test_flow_decorator.py
```diff
@@ -161,6 +152,10 @@ def test_flow_decorator_flow_in_flow(
     # WHEN Calling the _test_flow_in_flow function with specific messages
     _flow_over_flow(call_llm_messages)
 
+    # Wait for the Prompt span to be exported; It was asynchronously waiting
+    # on the OpenAI call span to finish first
+    time.sleep(1)
```
andreibratu, line 167: Added some time.sleep in tests since the processor now waits async before sending the HL spans to processor

tests/utilities/test_flow_decorator.py
```diff
@@ -281,18 +257,17 @@ def test_flow_decorator_hl_exporter_flow_inside_flow(
...
-        flow_span = mock_export_method.call_args_list[4][0][0][0]
-        nested_flow_span = mock_export_method.call_args_list[3][0][0][0]
-        assert (last_span_flow_metadata := TRACE_FLOW_CONTEXT.get(flow_span.get_span_context().span_id))
-        assert (flow_span_flow_metadata := TRACE_FLOW_CONTEXT.get(nested_flow_span.get_span_context().span_id))
-        assert flow_span_flow_metadata["trace_parent_id"] == flow_span.context.span_id
-        assert last_span_flow_metadata["is_flow_log"]
-        assert flow_span_flow_metadata["is_flow_log"]
+
+        # THEN one of the flows is nested inside the other
+        spans: list[ReadableSpan] = [mock_export_method.call_args_list[i][0][0][0] for i in range(1, 5)]
```
andreibratu, line 277: Order of spans is no longer guaranteed due to async mechanism in processor, tests should no longer rely or test for order

src/humanloop/otel/exporter.py
```diff
@@ -178,42 +176,64 @@ def _do_work(self):
...
+                    span_to_export.name,
                 )
             else:
-                # Requeue the Span to be uploaded later
+                # Requeue the Span and upload after its parent
                 self._upload_queue.put((span_to_export, evaluation_context))
             self._upload_queue.task_done()
 
+    def _mark_span_completed(self, span_id: int) -> None:
+        for flow_log_span_id, flow_children_span_ids in self._flow_log_prerequisites.items():
```
olehvell-h, line 203: potentially dumb question: Right now we mark flow as complete only if `self._flow_log_prerequisites.items()` is not empty - is it always the case? Aka do we ever need to mark it as complete when self._flow_log_prerequisites.items() is empty

src/humanloop/otel/exporter.py
```diff
@@ -178,42 +176,64 @@ def _do_work(self):
...
+        parent_span_id = span.parent.span_id if span.parent else None
+
+        while parent_span_id and self._span_id_to_uploaded_log_id.get(parent_span_id) is None:
+            logger.debug(
+                "[HumanloopSpanExporter] Span %s %s waiting for parent %s to be uploaded",
+                span.context.span_id,
+                span.name,
+                parent_span_id,
+            )
+            time.sleep(0.1)
```
olehvell-h: discussed offline - if time.sleep potentailly increases latency - maybe remove it

src/humanloop/prompts/client.py
```diff
@@ -3094,116 +3100,125 @@ async def upsert(
...
+                        template=[
+                            {
+                                "content": "You are a helpful coding assistant specialising in {{language}}",
+                                "role": "system",
+                            }
+                        ],
+                        provider="openai",
+                        max_tokens=-1,
+                        temperature=0.7,
+        <<<<<<< HEAD
```
olehvell-h, line 3208: oups

pyproject.toml
```diff
@@ -1,3 +1,6 @@
+[project]
+name = "humanloop"
```
olehvell-h, line 2: Is it necessary, aka why was it added now?

olehvell-h (approved):
make sense to me! 

I haven't propertly understood all logic, so don't count to much on my review. At the same time I don't see a reason not to approve it.   Maybe spend a bit of extra time on qaing just to make sure.

PR #45, opened 2025-02-24
Decorator fixes

andreibratu:
* Use /otel endpoint instead of processing spans locally
* Helpful error messages on evaluations.run()
* Throw errors for invalid decorator uses
* Decorators will no longer serialize all errors but raise critical ones -> evaluations will not hang mysteriously
* Move away from context variables in favor of OTEL's runtime context API
* @flow will also pick up SDK logging e.g. prompts.call(...) or tool.log(...)
* Sync with backend API: jinja release, log status use


andreibratu:
Test script
```python
from humanloop import Humanloop
from openai import OpenAI

hl_client = Humanloop(
    api_key="<HL_API_KEY>",
    base_url="http://localhost:80/v5",
)

TEMPLATE = "You are a useful chatbot that speaks like a {{personality}}"

client = OpenAI(api_key="<OPENAI_KEY>")
hl_client.prompts.populate_template(TEMPLATE, inputs={"personality": "pirate"})
messages = [
    {
        "role": "system",
        "content": "You are a useful chatbot that speaks like a ",
    }
]


@hl_client.tool(path="Andrei QA/Calculator")
def calculator(a: int, b: int):
    return a + b


@hl_client.flow(path="Andrei QA/Flow")
def fn(cool: str, beans: int):
    output = calculator(1, 2)
    hl_client.tools.log(
        path="Andrei QA/Log",
        tool={
            "function": {
                "name": "calculator",
                "description": "Adds two numbers",
                "parameters": {
                    "a": {
                        "type": "int",
                        "description": "First number",
                    },
                    "b": {
                        "type": "int",
                        "description": "Second number",
                    },
                },
            }
        },
        inputs={
            "a": 1,
            "b": 2,
        },
        output=output,
    )

    hl_client.prompts.call(
        path="Andrei QA/Call Prompt",
        prompt={
            "provider": "openai",
            "model": "gpt-4o",
            "temperature": 0.8,
            "frequency_penalty": 0.6,
            "presence_penalty": 0.6,
        },
        messages=[
            {
                "content": "Say something funny",
                "role": "system",
            }
        ],
    )

    with hl_client.prompt(path="Andrei QA/Prompt"):
        while True:
            user_input = input("> ")
            if user_input == "exit":
                break
            messages.append(
                {
                    "role": "user",
                    "content": user_input,
                }
            )
            response = (
                client.chat.completions.create(
                    model="gpt-4o",
                    messages=messages,
                    temperature=0.8,
                    frequency_penalty=0.6,
                    presence_penalty=0.6,
                    seed=42,
                )
                .choices[0]
                .message.content
            )
            print(response)
            messages.append(
                {
                    "role": "assistant",
                    "content": response,
                }
            )

    return messages[-1]


fn("cool", 1)

# @hl_client.flow(path="Andrei QA/Flow Evaluate")
# def fn_evaluate(a: int, b: int):
#     return calculator(a, b)


# hl_client.evaluations.run(
#     file={
#         "path": "Andrei QA/Flow Evaluate",
#         "callable": fn_evaluate,
#     },
#     name="Test",
#     dataset={
#         "path": "Andrei QA/Arithmetic",
#         "datapoints": [
#             {
#                 "inputs": {
#                     "a": 1,
#                     "b": 2,
#                 },
#                 "target": {
#                     "value": 3,
#                 },
#             },
#             {
#                 "inputs": {
#                     "a": 3,
#                     "b": 4,
#                 },
#                 "target": {
#                     "value": 7,
#                 },
#             },
#         ],
#     },
#     evaluators=[
#         {
#             "path": "Andrei QA/Equals",
#             "callable": lambda x, y: x["output"] == y["target"]["value"],
#             "return_type": "boolean",
#             "args_type": "target_required",
#         }
#     ],
# )
```

src/humanloop/client.py
```diff
@@ -136,22 +135,20 @@ def __init__(
...
         else:
             self._opentelemetry_tracer = opentelemetry_tracer
 
+    @contextmanager
     def prompt(
         self,
         *,
-        path: Optional[str] = None,
-        **prompt_kernel: Unpack[DecoratorPromptKernelRequestParams],  # type: ignore
+        path: str,
```
jamesbaskerville, line 154: nit for all of these: we should support passing either path or id, no?

src/humanloop/client.py
```diff
@@ -136,22 +135,20 @@ def __init__(
...
-            ),
+            HumanloopSpanProcessor(exporter=HumanloopSpanExporter(client=self)),
         )
 
         if opentelemetry_tracer is None:
             self._opentelemetry_tracer = self._tracer_provider.get_tracer("humanloop.sdk")
         else:
             self._opentelemetry_tracer = opentelemetry_tracer
 
+    @contextmanager
```
jamesbaskerville: as discussed earlier, this is cool but probably unnecessary -- simpler to just keep it as a decorator on a callable

src/humanloop/otel/exporter.py
```diff
@@ -128,265 +110,64 @@ def force_flush(self, timeout_millis: int = 10000) -> bool:
...
-        for Spans to arrive. Setting a timeout on self._upload_queue.get() risks
-        shutting down the thread early as no Spans are produced e.g. while waiting
-        for user input into the instrumented feature or application.
-
-        Each thread will upload a Span to Humanloop, provided the Span has all its
-        dependencies uploaded. The dependency happens in a Flow Trace context, where
-        the Trace parent must be uploaded first. The Span Processor will send in Spans
-        bottoms-up, while the upload of a Trace happens top-down. If a Span did not
-        have its span uploaded yet, it will be re-queued to be uploaded later.
-        """
```
jamesbaskerville, line 143: should still replace this with a correct docstring

peadaroh:
Just to confirm this refactors the 'polling behaviour' for flow logs to instead check for when it is updated.

src/humanloop/client.py
andreibratu: I'd keep it to path to reduce cognitive load

src/humanloop/client.py
andreibratu: yep done

src/humanloop/otel/exporter.py
andreibratu: outdated

src/humanloop/client.py
```diff
@@ -94,16 +98,14 @@ def __init__(
...
 
-        The Humanloop SDK File decorators use OpenTelemetry internally. You can provide a
-        TracerProvider and a Tracer if you'd like to integrate them with your existing
-        telemetry system. Otherwise, an internal TracerProvider will be used.
-        If you provide only the `TraceProvider`, the SDK will log under a Tracer
-        named `humanloop.sdk`.
+        The Humanloop SDK File decorators use OpenTelemetry internally.
+        You can provide a TracerProvider and a Tracer to integrate
+        with your existing telemetry system. If not provided,
+        an internal TracerProvider will be used.
```
jamesbaskerville, line 108: Does our OTEL exporting stuff still work ok if user passes in a TracerProvider?

src/humanloop/client.py
```diff
@@ -176,89 +167,87 @@ def call_llm(messages):
...
-
-        ```python
-        call_llm(messages=[
-            {"role": "system", "content": "You are an assistant on the following topics: finance."}
-            {"role": "user", "content": "What can you do?"}
-        ])
+        If a different model, endpoint, or hyperparameter is used, a new
+        Prompt version is created. For example:
         ```
+        @humanloop_client.prompt(path="My Prompt")
```
jamesbaskerville, line 186: nit: inconsistent, above example just uses `@prompt`. Should standardize

src/humanloop/client.py
```diff
@@ -176,89 +167,87 @@ def call_llm(messages):
...
-        If the function raises an exception, the log created by the function will have the output
-        field set to None and the error field set to the string representation of the exception.
+        And one Log will be added to each version of the Prompt.
+        ```
 
         :param path: The path where the Prompt is created. If not
             provided, the function name is used as the path and the File
             is created in the root of your Humanloop organization workspace.
 
         :param prompt_kernel: Attributes that define the Prompt. See `class:DecoratorPromptKernelRequestParams`
```
jamesbaskerville, line 224: looks like we're getting rid of this?

tests/integration/chat_agent/__init__.py
jamesbaskerville, line 1: are these tests all gone now? Or did they not work?

src/humanloop/overload.py
```diff
@@ -0,0 +1,124 @@
...
+        CreatePromptLogResponse,
+        CreateToolLogResponse,
+        CreateFlowLogResponse,
+        CreateEvaluatorLogResponse,
+    ]:
+        trace_id = get_trace_id()
+        if trace_id is not None and type(client) is FlowsClient:
+            context = get_decorator_context()
+            if context is None:
+                raise HumanloopRuntimeError("Internal error: trace_id context is set outside a decorator context.")
```
jamesbaskerville, line 54: if I'm understanding error right, this should also be a check in the line 59 block? i.e. apply even when not flow client.

I do like all this validation/error checking though!

src/humanloop/overload.py
```diff
@@ -0,0 +1,124 @@
...
+            }
+        evaluation_context = get_evaluation_context()
+        if evaluation_context is not None:
+            kwargs_eval, eval_callback = evaluation_context.log_args_with_context(
+                path=kwargs.get("path"), log_args=kwargs
+            )
+            try:
+                response = self._log(**kwargs_eval)
+            except Exception as e:
+                # Re-raising as HumanloopDecoratorError so the decorators don't catch it
```
jamesbaskerville, line 77: nice

src/humanloop/otel/processor.py
```diff
@@ -1,252 +1,54 @@
...
 
 logger = logging.getLogger("humanloop.sdk")
 
 
-class CompletableSpan(TypedDict):
-    span: ReadableSpan
-    complete: bool
-
-
 class HumanloopSpanProcessor(SimpleSpanProcessor):
```
jamesbaskerville, line 19: new docstring pls 🙏🏼

jamesbaskerville: the info about the instrumentors is still relevant, no?

src/humanloop/otel/processor.py
```diff
@@ -1,252 +1,54 @@
...
-        logger.debug("[HumanloopSpanProcessor] Sending span %s %s to exporter", span_id, span.name)
+    def on_start(self, span: Span, parent_context=...):
+        """Called when a Span is started."""
+        if is_llm_provider_call(span):
+            decorator_context = get_decorator_context()
+            if decorator_context and decorator_context.type == "prompt":
+                path, template = (
+                    decorator_context.path,
+                    decorator_context.version["template"],
+                )
```
jamesbaskerville, line 31: isn't path the only arg on prompt decorators now? or did I read that wrong and templates are still valid?

src/humanloop/otel/processor.py
```diff
@@ -1,252 +1,54 @@
...
+    def on_end(self, span: ReadableSpan):
+        """Called when a Span finishes recording."""
+        if is_llm_provider_call(span):
+            decorator_context = get_decorator_context()
+            if decorator_context is None or decorator_context.type != "prompt":
+                # User made a provider call outside a @prompt context, ignore the span
+                return
+            evaluation_context = get_evaluation_context()
+            if evaluation_context is not None and evaluation_context.path == decorator_context.path:
+                # User made a provider call inside a @prompt context, ignore the span
```
jamesbaskerville, line 52: fix: inside an evaluation context

also, is `evaluation_context.path` the prompt path being evaluated?

src/humanloop/otel/processor.py
```diff
@@ -1,252 +1,54 @@
...
+                return
         self.span_exporter.export([span])
```
jamesbaskerville, line 54: 😍

jamesbaskerville: the amount of redlines in this diff is amazing

src/humanloop/decorators/prompt.py
```diff
@@ -0,0 +1,44 @@
...
+    def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        @wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            with set_decorator_context(
+                DecoratorContext(
+                    path=path,
+                    type="prompt",
+                    version={
+                        # TODO: Implement a reverse-lookup of the template
+                        "template": None,
```
jamesbaskerville, line 26: btw is this a breaking change? Ok that template can't be specified in the prompt decorator (I think it is -- duplicate with prompts.call right?)

src/humanloop/decorators/tool.py
```diff
@@ -31,75 +32,90 @@
...
+        # Mypy complains about adding attribute on function, but it's nice DX
+        func.json_schema = tool_kernel["function"]  # type: ignore
 
         @wraps(func)
-        def wrapper(*args, **kwargs):
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> Optional[R]:
+            evaluation_context = get_evaluation_context()
+            if evaluation_context is not None:
+                if evaluation_context.path == path:
+                    raise HumanloopRuntimeError("Tools cannot be evaluated with the `evaluations.run()` utility.")
```
jamesbaskerville, line 63: Is this a TODO to support evaluating tools here?

src/humanloop/decorators/tool.py
```diff
@@ -31,75 +32,90 @@
...
-                span.set_attribute(HUMANLOOP_PATH_KEY, path if path else func.__name__)
-                span.set_attribute(HUMANLOOP_FILE_TYPE_KEY, "tool")
-                if enhanced_tool_kernel:
-                    write_to_opentelemetry_span(
-                        span=span,
-                        key=f"{HUMANLOOP_FILE_KEY}.tool",
-                        value=enhanced_tool_kernel,
-                    )
+                write_to_opentelemetry_span(
+                    span=span,  # type: ignore [arg-type]
```
jamesbaskerville, line 67: why was this a new type issue?

src/humanloop/decorators/tool.py
```diff
@@ -31,75 +32,90 @@
...
+                    key=HUMANLOOP_FILE_KEY,
+                    value=tool_kernel,  # type: ignore [arg-type]
+                )
+                span.set_attribute(HUMANLOOP_FILE_PATH_KEY, path)
```
jamesbaskerville, line 71: for myself -- equivalent to:
`write_to_opentelemetry_span(span=span, key=HUMANLOOP_FILE_PATH_KEY, value=path)`? And the helper just lets you easily write objects?

src/humanloop/decorators/flow.py
```diff
@@ -0,0 +1,133 @@
...
+                            func_output = None
+
+                        updated_flow_log = {
+                            "log_status": "complete",
+                            "output": log_output,
+                            "error": log_error,
+                            "output_message": log_output_message,
+                            "id": this_flow_log.id,
+                        }
+                        # Write the Flow Log to the Span on HL_LOG_OT_KEY
```
jamesbaskerville, line 115: typo: HUMANLOOP_LOG_KEY

src/humanloop/evals/run.py
```diff
@@ -245,39 +125,63 @@ def run_eval(
...
-        type=type_,
-    )
+    try:
+        hl_file = _upsert_file(file=file_, type=type_, client=client)
+    except ValidationError as e:
+        sys.stdout.write(f"{RED}Error in your `file` argument:\n\n{e}{RESET}")
+        return []
+    except Exception as e:
+        sys.stdout.write(f"{RED}Error in your `file` argument:\n\n{e}{RESET}")
+        return []
```
jamesbaskerville, line 145: nit: you can combine these like `except (ValidationError, Exception) as e`. Actually could just be Exception since you're not handling them differently

src/humanloop/evals/run.py
```diff
@@ -315,61 +219,82 @@ def upload_callback(log_id: str):
...
             for datapoint in hl_dataset.datapoints:
-                executor.submit(_process_datapoint, datapoint)
+                futures.append(executor.submit(_process_datapoint, datapoint))
+
+        for future in futures:
+            try:
+                future.result()
+            except Exception as e:
+                sys.stderr.write(f"\n{RED}Error processing datapoint:\n{_get_error_message(e)}{RESET}\n")
+                _cancel_evaluation()
```
jamesbaskerville, line 296: is it correct to cancel the whole eval when a single datapoint fails? seems excessive if I'm reading this right

src/humanloop/evals/run.py
```diff
@@ -467,10 +392,11 @@ def _wait_for_evaluation_to_complete(
             if not waiting_for_local_evals_message_printed:
                 sys.stderr.write("\n\nWaiting for Evaluators on Humanloop runtime...\n")
                 waiting_for_local_evals_message_printed = True
-            sys.stderr.write(stats.progress)
+            sys.stderr.write(stats.progress)  # type: ignore [arg-type]
             # Move the cursor up in stderr a number of lines equal to the number of lines in stats.progress
-            sys.stderr.write("\033[A" * (stats.progress.count("\n")))
+            # sys.stderr.write("\033[A" * (stats.progress.count("\n")))
             time.sleep(5)
```
jamesbaskerville, line 398: fix/followup: we did say to move away from this polling behavior. Prefer separate PR for it though

src/humanloop/evals/run.py
```diff
@@ -521,48 +447,63 @@ def _get_checks(
...
         # or more information than the `file` argument
         file_ = copy.deepcopy(inner_file)
+        if file_["type"] == "prompt":
+            sys.stdout.write(
+                f"{YELLOW}"
+                "The @prompt decorator will not spy on provider calls when passed to `evaluations.run()`. "
+                "Using the `version` in `file` argument instead.\n"
+                f"{RESET}"
+            )
+            # TODO: document this
```
jamesbaskerville, line 484: question: what does this mean?

src/humanloop/evals/run.py
```diff
@@ -893,6 +827,20 @@ def _check_evaluation_improvement(
...
+def _get_error_message(e: Exception, length_limit: bool = False) -> str:
+    import traceback
+
+    # Get the full traceback
+    trace_info = traceback.format_exc()
+
+    # Extract the last 200 characters of the traceback
+    last_trace_part = (
+        (trace_info[-200:] + "..." if len(trace_info) > 200 else trace_info) if length_limit else trace_info
+    )
```
jamesbaskerville, line 839: `...` should be in front if it's the last 200 chars, right?

jamesbaskerville: also feels like `length_limit` should just be an int instead of a boolean

src/humanloop/otel/exporter/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            span_to_export, eval_context_callback = thread_args
+
+            response = requests.post(
+                f"{self._client._client_wrapper.get_base_url()}/import/otel/v1/traces",
+                headers={
+                    **self._client._client_wrapper.get_headers(),
+                },
+                data=serialize_span(span_to_export),
+            )
+            print("RECV", span_to_export.attributes, response.json(), response.status_code)
```
jamesbaskerville, line 146: fix: change to logger/debug

src/humanloop/otel/exporter/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            if response.status_code != 200:
+                pass
+            else:
+                if eval_context_callback:
+                    print("HELLO")
```
jamesbaskerville, line 151: remove

src/humanloop/otel/exporter/proto.py
```diff
@@ -0,0 +1,76 @@
...
+                                events=[],
+                            )
+                        ],
+                    )
+                ]
+            )
+        ]
+    )
+
+    return MessageToJson(payload)
```
jamesbaskerville, line 76: hmm not 100% sure this will work as expected -- OTLP JSON encoding differs a bit from the proto3 standard mapping. Possible we should just be working in proto anyways? Might make things simpler since it's more of the default rather than JSON.

https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding

jamesbaskerville: I guess it's fine if our IDs aren't "correct" as long as they're internally consistent?

src/humanloop/otel/exporter/proto.py
```diff
@@ -0,0 +1,76 @@
...
+                                    2: ProtoBufferSpan.SpanKind.SPAN_KIND_CLIENT,
+                                    3: ProtoBufferSpan.SpanKind.SPAN_KIND_PRODUCER,
+                                    4: ProtoBufferSpan.SpanKind.SPAN_KIND_CONSUMER,
+                                }[span_to_export.kind.value],
+                                start_time_unix_nano=span_to_export.start_time,  # type: ignore [attr-defined, arg-type]
+                                end_time_unix_nano=span_to_export.end_time,  # type: ignore [attr-defined, arg-type]
+                                attributes=[
+                                    KeyValue(
+                                        key=key,
+                                        value=AnyValue(string_value=str(value)),
```
jamesbaskerville, line 46: are they always strings? Aren't they sometimes ints or some other type?

jamesbaskerville (commented):
generally looks great. Love the error/logging improvements and how simplified the processor/exporter logic is.

Few higher-level things:
- why are there more # type: ignore[arg-type] littered all over? what changed to cause that?
- need to handle the polling fix as a follow-on, though good we're sleeping in between polls now

jamesbaskerville (approved):

PR #48, opened 2025-02-28
Happy path fix

peadaroh:


src/humanloop/eval_utils/run.py
```diff
@@ -186,13 +199,19 @@ def _overload_log(
         try:
             response = self._log(**kwargs)
         except Exception as e:
-            logger.error(f"Failed to log: {e}")
+            error_message = str(e).replace("\n", " ")
+            if len(error_message) > 100:
```
peadaroh: This seems hacky, is it really needed?

src/humanloop/eval_utils/run.py
```diff
@@ -186,13 +199,19 @@ def _overload_log(
...
+                sys.stderr.write(f"{RED}Failed to log: {error_message[:100]}...{RESET}\n")
+            else:
+                sys.stderr.write(f"{RED}Failed to log: {error_message}{RESET}\n")
             raise e
 
         # Notify the run_eval utility about one Log being created
         if log_belongs_to_evaluated_file(log_args=kwargs):
             evaluation_context = get_evaluation_context()
+            evaluation_context.logged = True
```
peadaroh, line 213: is this for the specific datapoint and evaluation run?

src/humanloop/eval_utils/run.py
```diff
@@ -247,26 +266,40 @@ def run_eval(
...
     # Generate locally if a file `callable` is provided
     if function_ is None:
         # TODO: trigger run when updated API is available
-        logger.info(f"{CYAN}\nRunning '{hl_file.name}' over the Dataset '{hl_dataset.name}'{RESET}")
+        sys.stdout.write(f"{CYAN}\nRunning '{hl_file.name}' over the Dataset '{hl_dataset.name}'{RESET}\n")
     else:
         # Running the evaluation locally
-        logger.info(
-            f"{CYAN}\nRunning '{hl_file.name}' over the Dataset '{hl_dataset.name}' using {workers} workers{RESET} "
+        sys.stdout.write(
```
peadaroh, line 300: why sys vs logger?

src/humanloop/eval_utils/run.py
```diff
@@ -318,31 +356,46 @@ def upload_callback(log_id: str):
                     start_time=start_time,
                     end_time=datetime.now(),
                 )
-                logger.warning(
-                    msg=f"\nYour {hl_file.type}'s `callable` failed for Datapoint: {dp.id}. \n Error: {str(e)}"
-                )
+                error_message = str(e).replace("\n", " ")
```
peadaroh, line 364: this logic was repeated above, why is it needed?

src/humanloop/eval_utils/run.py
```diff
@@ -318,31 +356,46 @@ def upload_callback(log_id: str):
...
+                    sys.stderr.write(
+                        f"\n{RED}Your {hl_file.type}'s `callable` failed for Datapoint: {dp.id}. Error: {error_message[:100]}...{RESET}\n"
+                    )
+                else:
+                    sys.stderr.write(
+                        f"\n{RED}Your {hl_file.type}'s `callable` failed for Datapoint: {dp.id}. Error: {error_message}{RESET}\n"
+                    )
 
         with ThreadPoolExecutor(max_workers=workers) as executor:
+            futures = []
```
peadaroh: could you explain the move to futures?

src/humanloop/eval_utils/run.py
```diff
@@ -360,6 +413,9 @@ def __init__(self, total: int):
     def increment(self):
         """Increment the progress bar by one finished task."""
         with self._lock:
+            # NOTE: There is a deadlock here that needs further investigation
```
peadaroh: what does this deadlock cause? how important is it to resolve?

src/humanloop/eval_utils/run.py
```diff
@@ -407,15 +460,23 @@ def _wait_for_evaluation_to_complete(
 ):
     # Wait for the Evaluation to complete then print the results
     complete = False
+
+    wrote_explainer = False
```
peadaroh: Weirdly named variable, i have no idea what this could be for fromthe name

src/humanloop/eval_utils/run.py
```diff
@@ -846,45 +910,67 @@ def _run_local_evaluators(
...
-                judgement = eval_function(log_dict)
-
-            _ = client.evaluators.log(
-                version_id=local_evaluator.version_id,
-                parent_id=log_id,
-                judgment=judgement,
-                id=local_evaluator.id,
-                start_time=start_time,
-                end_time=datetime.now(),
+                log_dict = log
```
peadaroh, line 923: in general assignment like this may cause issues

src/humanloop/eval_utils/run.py
```diff
@@ -846,45 +910,67 @@ def _run_local_evaluators(
...
+            except Exception as e:
+                _ = client.evaluators.log(
+                    parent_id=log_id,
+                    id=local_evaluator.id,
+                    error=str(e),
+                    start_time=start_time,
+                    end_time=datetime.now(),
+                )
+                error_message = str(e).replace("\n", " ")
+                if len(error_message) > 100:
```
peadaroh: this logic is repeated twice above, why is it needed?

src/humanloop/flows/client.py
```diff
@@ -279,6 +282,118 @@ def log(
             raise ApiError(status_code=_response.status_code, body=_response.text)
         raise ApiError(status_code=_response.status_code, body=_response_json)
 
+    def update_log(
```
peadaroh: Was surprised by this change in the diff.

Was this just added?

src/humanloop/otel/exporter.py
```diff
@@ -116,13 +116,13 @@ def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
     def shutdown(self) -> None:
         self._shutdown = True
         for thread in self._threads:
-            thread.join()
+            thread.join(timeout=5000 / 1000)
```
peadaroh: 5000/1000 - this is weird

src/humanloop/eval_utils/run.py
andreibratu: Yes - every datapoint is ran in isolation from other threads, so only one log statement is accepted per evaluated file - run - datapoint

src/humanloop/eval_utils/run.py
andreibratu: logger doesn't interact with ANSI codes such as color, move cursor out of the box - will look into whether the logger can accept those

src/humanloop/eval_utils/run.py
andreibratu: was investigating some hanging threads - reverting the change

src/humanloop/eval_utils/run.py
andreibratu: can no longer reproduce it now - will remove the comment

src/humanloop/eval_utils/run.py
andreibratu: ack

src/humanloop/otel/exporter.py
andreibratu: ack

src/humanloop/flows/client.py
jamesbaskerville: This was a separate change from reordering the update_log path to be under/right after creating a flow log. Just part of fern autogenerating SDK; quick win came from talking w Jordan and Robin about agents DX last week.

andreibratu: not sure - reverted the file to the version on main branch, works as expected - branch likely started before updating some docstrings on response models last week

src/humanloop/otel/processor/prompts.py
```diff
@@ -1,4 +1,4 @@
-import deepdiff
+import deepdiff  # type: ignore [import]
```
jamesbaskerville, line 1: is this necessary? or did we lose types from the version downgrade?

src/humanloop/eval_utils/run.py
andreibratu: fern side issue: LogResponse is polymorphic on the backend, so it's likely not parsing the json into a pedantic object

src/humanloop/eval_utils/run.py
andreibratu: Noticed 502s tend to return a full HTML, polluting the screen - so I've condensed the error by removing the newlines and trimmed the length

src/humanloop/utilities/flow.py
```diff
@@ -57,10 +57,7 @@ def wrapper(*args: Sequence[Any], **kwargs: Mapping[str, Any]) -> Any:
                 except Exception as e:
                     logger.error(f"Error calling {func.__name__}: {e}")
                     output = None
-                    output_stringified = jsonify_if_not_string(
-                        func=func,
-                        output=None,
-                    )
+                    output_stringified = None
```
jamesbaskerville, line 60: is this change ok?  what's the purpose?

jamesbaskerville (approved):
generally lgtm

src/humanloop/utilities/flow.py
andreibratu: yes - output should be none when error is thrown, otherwise the backend complains we provided two output properties on a flow log (this constraint was relaxed in the /otel PR that's not merged yet)

EDIT: seems the reverse did not solve the error, I likely pushed the import deepdiff line without checking the lining previously

src/humanloop/otel/processor/prompts.py
andreibratu: good point - the downgrade is not required. reverted deepdiff to higher version, removing the need for the ignore

PR #51, opened 2025-03-07
Add sleep to OTEL exporters

jamesbaskerville:
Infinite polling the export queue without a sleep was causing threads to take up a full core of CPU. Simple sleep fixes that.

### DIFF
--- a/src/humanloop/otel/exporter/__init__.py
+++ b/src/humanloop/otel/exporter/__init__.py
@@ -1,5 +1,6 @@
 import logging
 
+import time
 import typing
 from queue import Empty as EmptyQueue
 from queue import Queue
@@ -129,7 +130,8 @@ class HumanloopSpanExporter(SpanExporter):
                 # Don't block or the thread will never be notified of the shutdown
                 thread_args = self._upload_queue.get(block=False)  # type: ignore
             except EmptyQueue:
-                # Wait for the another span to arrive
+                # Wait for another span to arrive
+                time.sleep(0.1)
                 continue
 
             span_to_export, eval_context_callback = thread_args


src/humanloop/otel/exporter.py
```diff
@@ -153,6 +154,7 @@ def _do_work(self):
                 )  # type: ignore
             except EmptyQueue:
                 # Wait for the another span to arrive
```
peadaroh: fix typo

src/humanloop/otel/exporter.py
```diff
@@ -153,6 +154,7 @@ def _do_work(self):
...
+                time.sleep(0.1)
```
peadaroh: worth making it slightly longer? don't totally understand in what context this code is being used

jamesbaskerville: Yeah it's fine to make it longer, too. There's a task in the queue whenever a span should be exported which approximately means every generation/evaluation/call. In an evaluations.run setting with up to 20 workers that could happen quite a few times a second, but would probably be spiky given they're usually dependent on provider calls.

(exporter has 4 workers by default I think)

andreibratu:
Can you merge in the release branch?

src/humanloop/otel/exporter.py
harry-humanloop: claude seems to give a range of 100-500ms as suitable generally when there's no need for realtime 
	•	Real-time systems: 10-50ms
	•	Interactive applications: 50-100ms
	•	Background processing: 100-500ms
	•	Batch processing: 500ms+

jamesbaskerville: mmk, leaving as is sounds reasonable in that case

PR #52, opened 2025-03-19
Clio support async functions

andreibratu:
* Allows decorators to wrap async functions - copy pasted the decorator wrapper; large portion of changes stem from here, will revisit in the future
* Relax OTEL dependency restrictions to make it compatible with PyAI
* Use singleton pattern for decorators' OTEL tracer init
* Beneficiaries: Clio (can use latest SDK version + decorate async functions), Tofu (? singleton + up to date libraries might fix their hanging client init, will reach back to them)


andreibratu:
Stale - adressed in release 0.8.40b6.

PR #61, opened 2025-05-01
p_sync Milestone 1 (Pull operation + CLI + call and load overloads)

AlePouroullis:
- Create sync client with support for pulling files from the remote humanloop to local workspace). Includes filters for environment and path (where path can specify specific file or directory)
- Create CLI with `pull` + `history` (shows last <=5 sync operations with some metadata) commands. Uses the [click](https://click.palletsprojects.com/en/stable/) library
- Overload `.call()` and `.log()` functions to use serialized contents of `.prompt`/`.agent` local files when `use_local_files` is set to `true` when initializing the humanloop client.
- Address edge cases like specifying path + environment or path + version_id: warn the user that it bypasses the local file and uses the file from remote instead.
- Cache files in memory after loading from disk using LRU cache
- Integration tests for pulling and overloading (though needs a fair bit more work)


jamesbaskerville:
think this needs a merge with master -- agents has gone in

pyproject.toml
```diff
@@ -1,13 +1,18 @@
 [project]
 name = "humanloop"
+description = "The Humanloop Python Library"
+authors = []
+keywords = ["ai", "machine-learning", "llm", "sdk", "humanloop"]
```
harry-humanloop: don't think machine-learning is appropriate. where is this shown?

AlePouroullis: These keywords aren't shown, only used for indexing and increasing discoverability. 
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#keywords
But perhaps machine learning would not be a good idea. Can just remove this altogether. The main thing was adding the description so it doesn't just say None below the package name
Screenshot 2025-05-02 at 10 51 46

PR #62, opened 2025-05-07
p_sync Milestone 1: Create Sync Client, implement pull operation, CLI, and overloads for log + call

AlePouroullis:
Superseded by [PR 69](https://github.com/humanloop/humanloop-python/pull/69)
p_sync/base is the auto-generated SDK with my latest backend changes from [here](https://github.com/humanloop/hl-alpha/pull/5012). Did this to make the PR focused on the custom code changes.

- Create sync client with support for pulling files from the remote humanloop to local workspace). Includes filters for environment and path (where path can specify specific file or directory)
- Create CLI with `pull` command. Uses the [click](https://click.palletsprojects.com/en/stable/) library
- Overload `.call()` and `.log()` functions to use serialized contents of `.prompt`/`.agent` local files when `use_local_files` is set to `true` when initializing the humanloop client.
- Address edge cases like specifying path + environment or path + version_id. I employ the principle of ["explicit is better than implicit"](https://peps.python.org/pep-0020/), i.e. raise error if user initializes the client with `use_local_files=True` and tries to reference path + remote-specific information (version_id/environment).
- Cache files in memory after loading from disk using LRU cache
- Integration tests for sync client, call and log overloading, CLI

### Prerequisites
Run this [backend branch](https://github.com/humanloop/hl-alpha/tree/p_sync/accept-serialized-files) locally 

## Set up the SDK
Clone this branch
`git clone --branch p_sync/pull git@github.com:humanloop/humanloop-python.git`

Install dependencies 
`poetry install`

Activate the venv
`poetry shell`

### Test the CLI
In the same directory as this cloned SDK
`humanloop pull --base-url http://localhost:80/v5`

In another project:
1. `poetry init` Use default configuration
2. In pyproject.toml modify python version constraints to `>=3.9,<4`
3. Add the SDK as a dependency. 
3a. Get the absolute path to the local SDK using `pwd`
3b. In your playground project run `poetry add --editable <path to SDK>` 
4. Then install peer dependencies `poetry install`
5. From the virtual environment (`poetry shell`) you can test the CLI `humanloop pull --base-url http://localhost:80/v5`

Note that you can remove `--base-url` entirely as the relevant backend changes are live in prod.


.fernignore
```diff
@@ -13,10 +13,12 @@ mypy.ini
 README.md
 src/humanloop/decorators
 src/humanloop/otel
+src/humanloop/sync
+src/humanloop/cli
 
 ## Tests
 
-tests/
+tests
```
andreibratu: only ignore our tests, fern auto-generates some

andreibratu: tests/integration are ours

pyproject.toml
```diff
@@ -1,13 +1,20 @@
...
+description = "The Humanloop Python Library"
+authors = []
 
+# This section is used by Poetry for development and building
+# The metadata here is used during development but not published to PyPI
 [tool.poetry]
 name = "humanloop"
-version = "0.8.36"
-description = ""
+version = "0.8.36b1"
```
andreibratu: Should be b2, b1 is the agents

pyproject.toml
```diff
@@ -86,7 +91,10 @@ plugins = ["pydantic.mypy"]
 [tool.ruff]
 line-length = 120
 
+[tool.poetry.scripts]
+humanloop = "humanloop.cli.__main__:cli"
```
andreibratu, line 96: integration test: use subprocess to check if cli works

andreibratu: check if exit code 0 when calling cli command; maybe check for exit code 1 when api key not provided

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+if not logger.hasHandlers():
+    logger.addHandler(console_handler)
+
+# Color constants
+SUCCESS_COLOR = "green"
+ERROR_COLOR = "red"
+INFO_COLOR = "blue"
+WARNING_COLOR = "yellow"
+
+MAX_FILES_TO_DISPLAY = 10
```
andreibratu: either show all of tehm, or pass a -q to silence

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+
+def get_client(api_key: Optional[str] = None, env_file: Optional[str] = None, base_url: Optional[str] = None) -> Humanloop:
+    """Get a Humanloop client instance."""
```
andreibratu: add in dosctring: exits with error if api key not found

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+        default=None,
+        hidden=True,
+    )
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        return f(*args, **kwargs)
+    return wrapper
+
+def handle_sync_errors(f: Callable) -> Callable:
+    """Decorator for handling sync operation errors."""
```
andreibratu: better docstring: write to console if a CLI operation fails

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    return wrapper
+
+@click.group(
+    help="Humanloop CLI for managing sync operations.",
+    context_settings={
+        "help_option_names": ["-h", "--help"],
+        "max_content_width": 100,
+    }
+)
+def cli():
```
andreibratu: better docstring on  why this is a `pass`: group itself takes no actions, it's here only to display the subgroup's help menu

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    """Humanloop CLI for managing sync operations."""
+    pass
+
+@cli.command()
+@click.option(
+    "--path",
+    "-p",
+    help="Path to pull (file or directory). If not provided, pulls everything. "
+    "To pull a specific file, ensure the extension for the file is included (e.g. .prompt or .agent). "
+    "To pull a directory, simply specify the path to the directory (e.g. abc/def to pull all files under abc/def and its subdirectories).",
```
andreibratu:
```suggestion
    help="Path in the Humanloop workspace to pull locally. You can pull an entire directory, e.g. /my/directory, "
    "or a specific file, e.g. /my/directory/summarize.prompt. When pulling a directory, all sub-directories will be "
    "included in the pull."
```

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    env_file: Optional[str], 
+    base_dir: str, 
+    base_url: Optional[str], 
+    verbose: bool
+):
+    """Pull prompt and agent files from Humanloop to your local filesystem.
+
+    \b
+    This command will:
+    1. Fetch prompt and agent files from your Humanloop workspace
```
andreibratu:
```suggestion
    1. Fetch serializable files from your Humanloop workspace.
```

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    {base_dir}/
+    ├── prompts/
+    │   ├── my_prompt.prompt
+    │   └── nested/
+    │       └── another_prompt.prompt
+    └── agents/
+        └── my_agent.agent
+
+    The operation will overwrite existing files with the latest version from Humanloop
+    but will not delete local files that don't exist in the remote workspace.
```
andreibratu:
```suggestion
  If a file exists both locally and in the Humanloop workspace, the local file will be overwritten in favour of the one
  deployed in the workspace environment. Files that only exists in local will not be affected.
```

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    client = get_client(api_key, env_file, base_url)
+    sync_client = SyncClient(client, base_dir=base_dir, log_level=logging.DEBUG if verbose else logging.WARNING)
+    
+    click.echo(click.style("Pulling files from Humanloop...", fg=INFO_COLOR))
+    click.echo(click.style(f"Path: {path or '(root)'}", fg=INFO_COLOR))
+    click.echo(click.style(f"Environment: {environment or '(default)'}", fg=INFO_COLOR))
+    
+    if verbose: 
+        # Don't use the spinner in verbose mode as the spinner and sync client logging compete
+        successful_files = sync_client.pull(path, environment)
```
andreibratu: remove spinner - per irl discussion, it's finnicky and doesn't add utility

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    else:
+        with progress_context("Pulling files..."):
+            successful_files = sync_client.pull(path, environment)
+    
+    # Get metadata about the operation
+    metadata = sync_client.metadata.get_last_operation()
```
andreibratu: follow-up with a loom

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    click.echo(click.style("Pulling files from Humanloop...", fg=INFO_COLOR))
+    click.echo(click.style(f"Path: {path or '(root)'}", fg=INFO_COLOR))
+    click.echo(click.style(f"Environment: {environment or '(default)'}", fg=INFO_COLOR))
+    
+    if verbose: 
+        # Don't use the spinner in verbose mode as the spinner and sync client logging compete
+        successful_files = sync_client.pull(path, environment)
+    else:
+        with progress_context("Pulling files..."):
+            successful_files = sync_client.pull(path, environment)
```
andreibratu: return succesful_files and error_files here instead of using metadata

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    
+    # Get metadata about the operation
+    metadata = sync_client.metadata.get_last_operation()
+    if metadata:
+        # Determine if the operation was successful based on failed_files
+        is_successful = not metadata.get('failed_files') and not metadata.get('error')
+        duration_color = SUCCESS_COLOR if is_successful else ERROR_COLOR
+        click.echo(click.style(f"Pull completed in {metadata['duration_ms']}ms", fg=duration_color))
+        
+        if metadata['successful_files']:
```
andreibratu: don't use metadata for this

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+            click.echo(click.style(f"\nSuccessfully pulled {len(metadata['successful_files'])} files:", fg=SUCCESS_COLOR))
+            
+            if verbose: 
+                for file in metadata['successful_files']:   
+                    click.echo(click.style(f"  ✓ {file}", fg=SUCCESS_COLOR))
+            else:
+                files_to_display = metadata['successful_files'][:MAX_FILES_TO_DISPLAY]
```
andreibratu: if len(failed_failes) > 0, cancel the operation, inform user via message in terminal, exit 1

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+
+@cli.command()
+@click.option(
+    "--oneline",
+    is_flag=True,
+    help="Display history in a single line per operation",
+)
+@handle_sync_errors
+@common_options
+def history(api_key: Optional[str], env_file: Optional[str], base_dir: str, base_url: Optional[str], oneline: bool):
```
andreibratu: needless:
- introduces state
- not in initial scope
- user can look into .humanloop to check the outcome of CLI command

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,256 @@
...
+    """Show sync operation history."""
+    client = get_client(api_key, env_file, base_url)
+    sync_client = SyncClient(client, base_dir=base_dir)
```
andreibratu: document somewhere that the default base_dir is `humanloop`

src/humanloop/cli/progress.py
```diff
@@ -0,0 +1,120 @@
+import sys
+import time
+from typing import Optional, Callable, Any
+from threading import Thread, Event
+from contextlib import contextmanager
+
+class Spinner:
```
andreibratu: remove module + spinner

src/humanloop/overload.py
```diff
@@ -114,11 +118,105 @@ def _overload_call(self, **kwargs) -> PromptCallResponse:
...
+    def _overload(self, function_name: str, **kwargs) -> PromptCallResponse:
+        if "id" in kwargs and "path" in kwargs: 
+            raise HumanloopRuntimeError(f"Can only specify one of `id` or `path` when {function_name}ing a {file_type}")
+        # Handle local files if enabled
+        if use_local_files and "path" in kwargs:
+            # Check if version_id or environment is specified
+            has_version_info = "version_id" in kwargs or "environment" in kwargs
+            normalized_path = sync_client._normalize_path(kwargs["path"])
+            
+            if has_version_info:
```
andreibratu: better name: use_workspace = any(["version_id" in kwargs, "environment" in kwargs])

src/humanloop/overload.py
```diff
@@ -114,11 +118,105 @@ def _overload_call(self, **kwargs) -> PromptCallResponse:
...
+                logger.warning(
+                    f"Ignoring local file for `{normalized_path}` as version_id or environment was specified. "
+                    "Using remote version instead."
+                )
+            else:
+                # Only use local file if no version info is specified
+                try:
+                    # If file_type is already specified in kwargs, it means user provided a PromptKernelRequestParams object
+                    if file_type in kwargs and not isinstance(kwargs[file_type], str):
+                        logger.warning(
```
andreibratu: throw error here

src/humanloop/sync/metadata_handler.py
```diff
@@ -0,0 +1,125 @@
+import json
+import time
+from datetime import datetime
+from pathlib import Path
+from typing import Dict, List, Optional, TypedDict, NotRequired
+import logging
+
+logger = logging.getLogger(__name__)
```
andreibratu: remove metadata, keep cli stateless

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+console_handler = logging.StreamHandler()
+formatter = logging.Formatter("%(message)s")
+console_handler.setFormatter(formatter)
+if not logger.hasHandlers():
+    logger.addHandler(console_handler)
+
+# Default cache size for file content caching
+DEFAULT_CACHE_SIZE = 100
+
+def format_api_error(error: Exception) -> str:
```
andreibratu, line 26: Like this, in future release refactor to a reusable function

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+        client: Humanloop client instance
+        base_dir: Base directory for synced files (default: "humanloop")
+        cache_size: Maximum number of files to cache (default: DEFAULT_CACHE_SIZE)
+        log_level: Log level for logging (default: WARNING)
+        """
+        self.client = client
+        self.base_dir = Path(base_dir)
+        self._cache_size = cache_size
+
+        global logger
```
andreibratu: do we need the global here? we are not writing to logger, read only from global should be fine

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+        
+        Args:
+            path: The normalized path to the file (without extension)
+            file_type: The type of file (Prompt or Agent)
+            
+        Returns:
+            The raw file content
+            
+        Raises:
+            HumanloopRuntimeError: If the file doesn't exist or can't be read
```
andreibratu, line 149:
```suggestion
            HumanloopRuntimeError: File doesn't exist or the read failed
```

// what does read failed mean? elaborate: disk write errors e.g.

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+
+            # Add file type extension
+            new_path = full_path.parent / f"{full_path.stem}.{file_type}"
+
+            # Write raw file content to file
+            with open(new_path, "w") as f:
+                f.write(serialized_content)
+            
+            # Clear the cache for this file to ensure we get fresh content next time
+            self.clear_cache()
```
andreibratu: Do we need the cache? expect the GET request to be fast since client is offloading all processing to us

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                response = self.client.files.list_files(
+                    type=["prompt", "agent"], 
+                    page=page,
+                    include_raw_file_content=True,
+                    environment=environment,
+                    path=path
+                )
+
+                if len(response.records) == 0:
+                    logger.debug("No more files found")
```
andreibratu:
```suggestion
                    logger.debug(f"Finished reading files for path {path}")
```

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                    break
+
+                logger.debug(f"Found {len(response.records)} files from page {page}")
```
andreibratu:
```suggestion
                logger.debug(f"{path}: Read page {page_idx} containing {len(records)} files")
```

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                        logger.warning(f"Skipping unsupported file type: {file.type}")
+                        continue
+
+                    # Skip if no raw file content
+                    if not getattr(file, "raw_file_content", None):
+                        logger.warning(f"No content found for {file.type} {getattr(file, 'id', '<unknown>')}")
+                        continue
+
+                    try:
+                        logger.debug(f"Saving {file.type} {file.path}")
```
andreibratu:
```suggestion
                        logger.debug(f"Writing {file.type} {file.path} to disk")
```

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+        Raises:
+            HumanloopRuntimeError: If the file doesn't exist or can't be read
+        """
+        return self._get_file_content_impl(path, file_type)
+
+    def clear_cache(self) -> None:
+        """Clear the LRU cache."""
+        self.get_file_content.cache_clear()
+
+    def _normalize_path(self, path: str) -> str:
```
andreibratu, line 157: Should throw for 2 and 5

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+            path: The path to normalize
+
+        Returns:
+            The normalized path
+        """
+        # Remove any file extensions
+        path = path.rsplit('.', 1)[0] if '.' in path else path
+        
+        # Convert backslashes to forward slashes and normalize multiple slashes
+        path = path.replace('\\', '/')
```
andreibratu: what about '\\\\'

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+        # Remove leading/trailing whitespace and slashes
+        path = path.strip().strip('/')
+        
+        # Normalize multiple consecutive slashes into a single forward slash
+        while '//' in path:
+            path = path.replace('//', '/')
+            
+        return path
+
+    def is_file(self, path: str) -> bool:
```
andreibratu: fragile, use a regex like `r'\.\w+$'`

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                if len(response.records) == 0:
+                    logger.debug("No more files found")
+                    break
+
+                logger.debug(f"Found {len(response.records)} files from page {page}")
+
+                # Process each file
+                for file in response.records:
+                    # Skip if not a Prompt or Agent
+                    if file.type not in ["prompt", "agent"]:
```
andreibratu: move this to a class attribute "serializible files" - even better, use `FileType` enum

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+        logger.info(f"Starting pull operation: path={normalized_path or '(root)'}, environment={environment or '(default)'}")
+        try:
+            if path is None:
+                # Pull all files from the root
+                logger.debug("Pulling all files from root")
+                successful_files = self._pull_directory(None, environment)
+                failed_files = []  # Failed files are already logged in _pull_directory
+            else:
+                if self.is_file(path.strip()):
+                    logger.debug(f"Pulling specific file: {normalized_path}")
```
andreibratu:
```suggestion
                    logger.debug(f"Pulling file: {normalized_path}")
```

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                    successful_files = self._pull_directory(normalized_path, environment)
+                    failed_files = []  # Failed files are already logged in _pull_directory
+
+            duration_ms = int((time.time() - start_time) * 1000)
+            logger.info(f"Pull completed in {duration_ms}ms: {len(successful_files)} files succeeded")
+
+            # Log the successful operation
+            self.metadata.log_operation(
+                operation_type="pull",
+                path=normalized_path or "",  # Use empty string if path is None
```
andreibratu: have `""` be a default argument and mention the behaviour in the docstring

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,327 @@
...
+                successful_files=successful_files,
+                failed_files=failed_files,
+                duration_ms=duration_ms
+            )
+            
+            return successful_files
+        except Exception as e:
+            duration_ms = int((time.time() - start_time) * 1000)
+            # Log the failed operation
+            self.metadata.log_operation(
```
andreibratu: would argue for not keeping track of duration; if necessary, return it from function alongside succesful_files and error_files

tests/sync/test_sync.py
```diff
@@ -0,0 +1,267 @@
...
+            model=file.model, 
+            id=response.id, 
+            version_id=response.version_id
+        ))
+
+    return created_files
+
+
+@pytest.fixture
+def cleanup_local_files():
```
andreibratu: Nice - eventually add the cleanup logic after the yield in the fixture that creates the files. E.g.

This can be your test_file_structure
```python
@pytest.fixture
def create_local_test_dir()
     # ... file creation goes here
     yield list_of_files
     # ... cleanup logic goes here
```

andreibratu: just fyi, autouse arg exists on @pytest.fixture

tests/sync/test_sync.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # Basic assertions
+        assert local_path.exists(), f"Expected synced file at {local_path}"
+        assert local_path.parent.exists(), f"Expected directory at {local_path.parent}"
+
+        # Verify it's not empty
+        content = local_path.read_text()
+        assert content, f"File at {local_path} should not be empty"
+
+@pytest.mark.parametrize("humanloop_client", [True], indirect=True)
```
andreibratu: indirect? also only one value, still need parameterize?

andreibratu (changes requested):
Remove the flashy parts, don't think the cache is necessary, have small nits. Integration tests for CLI would also be useful

src/humanloop/cli/__main__.py
AlePouroullis: I've decided against this as I don't think users will understand what 'serializable' means.

src/humanloop/cli/__main__.py
AlePouroullis: The problem I'm seeing with this is it would make things quite complex to support transactions, i.e. allow rollbacks if one file fails.

tests/sync/test_sync.py
AlePouroullis: `indirect=True` passes the parameter value to the fixture instead of the test function. In our case, it lets us configure `use_local_files=True` in the `humanloop_client` fixture.

tests/sync/test_sync.py
AlePouroullis: Thanks for the info 🙌

src/humanloop/sync/sync_client.py
AlePouroullis: I've removed the metadata handler entirely for now. I still log the duration when successful though

src/humanloop/sync/sync_client.py
AlePouroullis: Curious to know the rationale here? I feel None is more explicit nothing being specified. Also, I've tried using the list files endpoint when specifying empty string for path and it fails with the following error
`Directory `\"\"` does not exist.`
It seems the way I set it up doesn't gracefully handle empty strings. I'll have to open up another PR to address this. We could just convert empty string to None when making the actual API request

src/humanloop/sync/sync_client.py
AlePouroullis: FileType is Union of literals (and typing.Any for some reason?). I've created a class attribute SERIALIZABLE_FILE_TYPES which constraints the supported file types.

src/humanloop/sync/sync_client.py
AlePouroullis: This cache is used during runtime, so that we only need to read from disk once, then can get from memory.

src/humanloop/sync/sync_client.py
AlePouroullis: Great catch. Addressed 👍

src/humanloop/sync/sync_client.py
AlePouroullis: I now use pathlib instead, which is much more robust
  ```  
  def _normalize_path(self, path: str) -> str:
        """Normalize the path by:
        1. Converting to a Path object to handle platform-specific separators
        2. Removing any file extensions
        3. Converting to a string with forward slashes and no leading/trailing slashes
        """
        # Convert to Path object to handle platform-specific separators
        path_obj = Path(path)
        
        # Remove extension, convert to string with forward slashes, and remove leading/trailing slashes
        normalized = str(path_obj.with_suffix(''))
        # Replace all backslashes and normalize multiple forward slashes
        return '/'.join(part for part in normalized.replace('\\', '/').split('/') if part)

src/humanloop/sync/sync_client.py
AlePouroullis: Good point. Resolved

src/humanloop/overload.py
AlePouroullis: I error on version_id/environment being specified, but not when prompt/agent config specified, because that doesn't conflict with the notion of using local files.

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,206 @@
...
+            load_dotenv(env_file)
+        else:
+            env_path = find_dotenv()
+            if env_path:
+                load_dotenv(env_path)
+            else:
+                if os.path.exists(".env"):
+                    load_dotenv(".env")
+                else:
+                    load_dotenv()
```
jamesbaskerville: nit: think this can be simplified to `load_dotenv(find_dotenv())` from what I can read

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,206 @@
...
+def cli(): # Does nothing because used as a group for other subcommands (pull, push, etc.)
+    """Humanloop CLI for managing sync operations."""
+    pass
+
+@cli.command()
+@click.option(
+    "--path",
+    "-p",
+    help="Path in the Humanloop workspace to pull from (file or directory). You can pull an entire directory (e.g. 'my/directory/') "
+    "or a specific file (e.g. 'my/directory/my_prompt.prompt'). When pulling a directory, all files within that directory and its subdirectories will be included.",
```
jamesbaskerville: Specify default is workspace root

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,206 @@
...
+    base_url: Optional[str], 
+    verbose: bool,
+    quiet: bool
+):
+    """Pull Prompt and Agent files from Humanloop to your local filesystem.
+
+    \b
+    This command will:
+    1. Fetch Prompt and Agent files from your Humanloop workspace
+    2. Save them to your local filesystem (default: humanloop/)
```
jamesbaskerville:
```suggestion
    2. Save them to your local filesystem (default directory: humanloop/)
```

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,372 @@
...
+    error_msg = str(error)
+    if "status_code" not in error_msg or "body" not in error_msg:
+        return error_msg
+
+    try:
+        # Extract the body part and parse as JSON
+        body_str = error_msg.split("body: ")[1]
+        # Convert Python dict string to valid JSON by replacing single quotes with double quotes
+        body_str = body_str.replace("'", '"')
+        body = json.loads(body_str)
```
jamesbaskerville: I think you might first need to escape any existing double quotes in the dict string

src/humanloop/overload.py
```diff
@@ -112,13 +120,130 @@ def _overload_call(self, **kwargs) -> PromptCallResponse:
...
+
+    Raises:
+        HumanloopRuntimeError: If use_local_files is True and local file cannot be accessed
+    """
+    if not hasattr(client, "_overloads"):
+        client._overloads = defaultdict(list)  # type: ignore [union-attr]
+    # If the method has been overloaded, don't re-add the method
+    if isinstance(client, PromptsClient):
+        if len(client._overloads["call"]) == 1:  # type: ignore [attr-defined]
+            client._overloads["call"].append(client.call)  # type: ignore [attr-defined]
```
jamesbaskerville: why is this 1 while the others check against 0? Would appreciate extra comment as fairly unintuitive for someone on the outside

src/humanloop/overload.py
```diff
@@ -112,13 +120,130 @@ def _overload_call(self, **kwargs) -> PromptCallResponse:
...
+    else:
+        raise NotImplementedError(f"Unsupported client type: {type(client)}")
+    if len(client._overloads["log"]) == 0:
+        client._overloads["log"].append(client.log)
+    else:
+        raise RuntimeError(f"Unexpected overload order of operations for {client}.log")
+
+    file_type = _get_file_type_from_client(client)
+
+    def _overload(self, function_name: str, **kwargs) -> PromptCallResponse:
```
jamesbaskerville: nit: presumably return type shouldn't always be `PromptCallResponse`

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,372 @@
...
+            path: The normalized path to the file (without extension)
+            file_type: The type of file (Prompt or Agent)
+
+        Returns:
+            The raw file content
+
+        Raises:
+            HumanloopRuntimeError: If the file doesn't exist or can't be read
+        """
+        return self._get_file_content_implementation(path, file_type)
```
jamesbaskerville, line 151: fix: Shouldn't this be the cached version `self.get_file_content`?

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+            # Add file type extension
+            new_path = full_path.parent / f"{full_path.stem}.{file_type}"
+
+            # Write raw file content to file
+            with open(new_path, "w") as f:
+                f.write(serialized_content)
+
+            # Clear the cache when a file is saved
+            self.clear_cache()
```
jamesbaskerville: do we have to clear the cache every time we save a file? surprising. Would kinda expect that we just do this once after finishing the pull

src/humanloop/sync/sync_client.py
```diff
@@ -0,0 +1,372 @@
...
+        """Clear the LRU cache."""
+        # @TODO: @ale, why not put the cache on the client? This is a bit of a hack.
+        self.get_file_content.cache_clear()  # type: ignore [attr-defined]
+
+    def _normalize_path(self, path: str) -> str:
+        """Normalize the path by:
+        1. Converting to a Path object to handle platform-specific separators
+        2. Removing any file extensions
+        3. Converting to a string with forward slashes and no leading/trailing slashes
+        """
```
jamesbaskerville, line 162: Just clarifying -- this is for local filesystem path -> humanloop workspace path?

tests/custom/conftest.py
```diff
@@ -87,10 +79,25 @@ def opentelemetry_test_configuration(
...
+@pytest.fixture(scope="session")
+def get_humanloop_client() -> GetHumanloopClientFn:
+    load_dotenv()
+    if not os.getenv("HUMANLOOP_API_KEY"):
+        pytest.fail("HUMANLOOP_API_KEY is not set for integration tests")
+
+    def _get_humanloop_test_client(use_local_files: bool = False) -> Humanloop:
+        return Humanloop(
+            api_key=os.getenv("HUMANLOOP_API_KEY"),
+            base_url="http://localhost:80/v5",
```
jamesbaskerville: surprised at this -- would expect tests to hit our prod or staging APIs

edit: or, more likely, be configured with an env var

jamesbaskerville (commented):
not gonna look at tests too closely yet, but looking good so far!

src/humanloop/overload.py
AlePouroullis: Noted. Do you mind taking a look at overload.py again please? I've done a large overhaul. There were a couple issues and some duplication.

src/humanloop/overload.py
AlePouroullis: Fair point. As mentioned before, I've completely overhauled for this reason: didn't feel particularly intuitive.

olehvell-h:
SDK relies on relative path? 
```
humanloop.prompts.call(path="/Users/oleh/humanloop/test-sync-sdk/humanloop-python/humanloop/Oleh's space/Customer Support bot.prompt")
```
Local file not found: humanloop/Users/oleh/humanloop/test-sync-sdk/humanloop-python/humanloop/Oleh's space/Customer Support bot.prompt


pwd of the file I'm calling the SDK from
```
/Users/oleh/humanloop/playground/test-sdk
```


I haven't thought of the implication of it.

olehvell-h:
> SDK relies on relative path?
> 
> ```
> humanloop.prompts.call(path="/Users/oleh/humanloop/test-sync-sdk/humanloop-python/humanloop/Oleh's space/Customer Support bot.prompt")
> ```
> 
> Local file not found: humanloop/Users/oleh/humanloop/test-sync-sdk/humanloop-python/humanloop/Oleh's space/Customer Support bot.prompt
> 
> pwd of the file I'm calling the SDK from
> 
> ```
> /Users/oleh/humanloop/playground/test-sdk
> ```
> 
> I haven't thought of the implication of it.

I see, I answered my own question. I thought path means my localpath to .prompt file.  This is a DX gotcha

olehvell-h:
```
humanloop.core.api_error.ApiError: status_code: 401, body: {'detail': 'Could not validate credentials.'}

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/oleh/humanloop/playground/test-sdk/letscallhllocally.py", line 9, in <module>
    xx = humanloop.prompts.call(path="/Oleh's space/Customer Support bot.prompt")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/oleh/humanloop/test-sync-sdk/humanloop-python/src/humanloop/overload.py", line 204, in call_wrapper
    return _overload_call(self, sync_client, use_local_files, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/oleh/humanloop/test-sync-sdk/humanloop-python/src/humanloop/overload.py", line 175, in _overload_call
    raise HumanloopRuntimeError from e
humanloop.error.HumanloopRuntimeError
```

I think there is an error that we don't handle properly. Would expect to see "humanloop.core.api_error.ApiError: status_code: 401, body: {'detail': 'Could not validate credentials.'}" as first line, not
```
....
  File "/Users/oleh/humanloop/test-sync-sdk/humanloop-python/src/humanloop/overload.py", line 204, in call_wrapper
    return _overload_call(self, sync_client, use_local_files, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/oleh/humanloop/test-sync-sdk/humanloop-python/src/humanloop/overload.py", line 175, in _overload_call
    raise HumanloopRuntimeError from e
humanloop.error.HumanloopRuntimeError
```

olehvell-h:
image

I made a typo and new file was created on HL.  Curious whether it makes sense to have strict mode, where creating new files (or even versions) is not allowed if `use_local_files=True` . Don't want to expand scope, just sharing my observations

src/humanloop/client.py
```diff
@@ -111,6 +116,21 @@ def __init__(
...
+        base_url: Optional base URL for the API
+        environment: The environment to use (default: DEFAULT)
+        api_key: Your Humanloop API key (default: from HUMANLOOP_API_KEY env var)
+        timeout: Optional timeout for API requests
+        follow_redirects: Whether to follow redirects
+        httpx_client: Optional custom httpx client
+        opentelemetry_tracer_provider: Optional tracer provider for telemetry
+        opentelemetry_tracer: Optional tracer for telemetry
+        use_local_files: Whether to use local files for prompts and agents
+        files_directory: Directory for local files (default: "humanloop")
```
olehvell-h: I think docstring should emphasize path to humanloop dir, not name of the dir itself. My usecase was

```
humanloop
nested_folder
--- nested_file.py
```

Inside nested_file I want to specify _path_  to humanloop dir.  files_directory does this job, but the docstring is not explicit about it

.gitignore
```diff
@@ -5,5 +5,7 @@ poetry.toml
 .ruff_cache/
 .vscode
 .env
+.env.test
 tests/assets/*.jsonl
 tests/assets/*.parquet
+humanloop
```
andreibratu, line 11: do add a comment explaining why 'humanloop' and '.env.test' are needed - speaking of, why do we need both .env and .env.test

pyproject.toml
```diff
@@ -1,13 +1,20 @@
...
+authors = []
 
+# This section is used by Poetry for development and building
+# The metadata here is used during development but not published to PyPI
 [tool.poetry]
 name = "humanloop"
-version = "0.8.36"
-description = ""
+version = "0.8.39"
+description = "Humanloop Python SDK"
```
andreibratu, line 12: have you changes this? this is auto-generated by fen

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
```
andreibratu, line 1: note: i have updated humanloop-docs with the pytest-xdist dependency; we'll regenerate the sdk and rebase this PR on top of it before merging

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Raises:
+        click.ClickException: If no API key is found
+    """
+    # Try loading from .env file
+    if env_file:
+        load_dotenv(env_file)
+    else:
+        # Try to find .env file in current directory or parent directories
+        env_path = find_dotenv()
```
andreibratu: why seraching in parents? this can have weird side effects in user's code; I'd try .env from current dir, or the one specified by user, and throw if not found

andreibratu: i'd even consider adding an --api_key flag, or defaulting to .env in project dir.

--api_key is better for CI workflows, and I think looking .env by custom path adds no utility

src/humanloop/cli/__main__.py
```diff
@@ -0,0 +1,243 @@
...
+            )
+        )
+
+    return api_key
+
+
+def get_client(
+    api_key: Optional[str] = None, env_file: Optional[str] = None, base_url: Optional[str] = None
+) -> Humanloop:
+    """Get a Humanloop client instance.
```
andreibratu:
```suggestion
    """Instantiate Humanloop client for CLI.
```

src/humanloop/client.py
andreibratu: agree with Oleh, rename to something like `local_directory`

tests/custom/conftest.py
AlePouroullis: Yeah, agreed. Left this in while I was doing tests to make sure I didn't screw with the prod files. But have reverted now

src/humanloop/client.py
andreibratu:
```suggestion
        local_directory: Local path for storing the serialised Humanloop Files
```

src/humanloop/client.py
```diff
@@ -111,6 +116,21 @@ def __init__(
...
+        cache_size: Maximum number of files to cache when use_local_files is True (default: DEFAULT_CACHE_SIZE).
```
andreibratu, line 141: would be good to throw or warn of use_local_files == False and cache_size is specified

andreibratu: also, any reason for providing users control to this? feels it adds complexity with no functional gain

src/humanloop/sync/sync_client.py
AlePouroullis: This path normalization is for specifying path in call/log or via path filter in cli, the former being the path specified in prompts.call, the latter for files by path in remote.

src/humanloop/sync/sync_client.py
AlePouroullis: Good point. Will change to what you've suggested

src/humanloop/files/raw_client.py
```diff
@@ -132,7 +132,7 @@ def list_files(
     def retrieve_by_path(
         self,
         *,
-        path: str,
+        path: str | None = None,
```
andreibratu: reminder: we modified auto-generated files, remove this

src/humanloop/files/client.py
```diff
@@ -118,7 +118,8 @@ def list_files(
     def retrieve_by_path(
         self,
         *,
-        path: str,
+        # @TODO: @ale, can this be None?
+        path: str | None = None,
```
andreibratu: reminder: we modified auto-generated files, remove this

tests/custom/conftest.py
andreibratu: @AlePouroullis this fixture is outdated right? in favour of `test_humanloop_client` - remove this and all other old fixtures

src/humanloop/cli/__main__.py
AlePouroullis: realized that find_dotenv() is actually used by load_dotenv() internally anyways. Though I decided to explicitly check if load_dotenv(env_file) returns False, which indicates that it failed to read from the specified file, which I then raise an error on.

src/humanloop/sync/sync_client.py
AlePouroullis: Good catch. Addressed

src/humanloop/sync/sync_client.py
AlePouroullis: The cached version is created during initialization (line 96)
`self.get_file_content = lru_cache(maxsize=cachesize)(self._get_file_content_implementation)`

src/humanloop/cli/__main__.py
AlePouroullis: Good point. I've simplified this function quite a bit.

src/humanloop/client.py
AlePouroullis: I disagree. `local_directory` is too ambiguous on what it stores and could be confused for storing any sort of data locally. 
Perhaps something like `local_files_directory` would be the most clear about purpose

pytest.ini
AlePouroullis: Thanks 🙏

src/humanloop/client.py
AlePouroullis: Good point. Just a bit tricky as cache_size is provided with a default int value, not None. Which means either we default to None and set it in the constructor if use_local_files is True so that we can check if a value is provided when use_local_files is False, or we just check if user set custom cache_size that differs from default.

pyproject.toml
AlePouroullis: This file's autogenerated by fern? Oops, thought we'd be able to modify project metadata.

PR #63, opened 2025-05-09
fix: Hotfix streaming for prompts/agents

jamesbaskerville:
Fixes regression to prompts.call streaming. Basically there was a refactor moving logic to `raw_client` files. In the process the streaming logic was changed and the method now does a `yield _sse.data()` where `_sse.data` is actually a string. This results in the enclosing method not actually yielding anything, so our streaming methods were very broken.

Simple hotfix here to remove the calling parens and yield the data string directly.

Confirmed working locally.

### DIFF
--- a/src/humanloop/agents/raw_client.py
+++ b/src/humanloop/agents/raw_client.py
@@ -578,7 +578,7 @@ class RawAgentsClient:
                                 if _sse.data == None:
                                     return
                                 try:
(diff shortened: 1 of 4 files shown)


harry-humanloop (approved):

jamesbaskerville:
fern bumped python generator and merged that instead

PR #70, opened 2025-05-15
Improve type annotations for overloaded client methods

AlePouroullis:
This PR fixes the type annotations in the client overloading mechanism to preserve method return types. When users call methods like client.prompts.call(), the type system now correctly identifies the return type as PromptCallResponse instead of Any.
The implementation uses getattr/setattr to avoid type errors while maintaining identical runtime behavior.

### DIFF
--- a/src/humanloop/overload.py
+++ b/src/humanloop/overload.py
@@ -1,7 +1,7 @@
 import inspect
 import logging
 import types
-from typing import Any, Callable, Dict, Optional, Union, TypeVar, Protocol
+from typing import Any, Callable, Dict, Optional, TypeVar, Union
 
 from humanloop.agents.client import AgentsClient
 from humanloop.context import (
@@ -59,13 +59,13 @@ def _get_file_type_from_client(
         return "dataset"
     elif isinstance(client, EvaluatorsClient):
         return "evaluator"
+    else:
+        raise ValueError(f"Unsupported client type: {type(client)}")
 
-    raise ValueError(f"Unsupported client type: {type(client)}")
 
-
-def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, Any]:
+def _handle_tracing_context(kwargs: Dict[str, Any], client: T) -> Dict[str, Any]:
     """Handle tracing context for both log and call methods."""
-    trace_id = get_trace_id()
+    trace_id = get_trace_id()   
     if trace_id is not None:
         if "flow" in str(type(client).__name__).lower():
             context = get_decorator_context()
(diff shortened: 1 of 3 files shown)


src/humanloop/overload.py
```diff
@@ -184,33 +188,39 @@ def _overload_call(self: Any, sync_client: Optional[SyncClient], use_local_files
...
     sync_client: Optional[SyncClient] = None,
     use_local_files: bool = False,
-) -> Any:
+) -> T:
     """Overloads client methods to add tracing, local file handling, and evaluation context."""
     # Store original log method as _log for all clients. Used in flow decorator
     if hasattr(client, "log") and not hasattr(client, "_log"):
-        client._log = client.log  # type: ignore[attr-defined]
+        # Store original method - using getattr/setattr to avoid type errors
+        original_log = getattr(client, "log")
```
peadaroh: I don't really understand why this is needed. How come it was not needed before? Could you explain why doing this avoids type errors

src/humanloop/overload.py
```diff
@@ -184,33 +188,39 @@ def _overload_call(self: Any, sync_client: Optional[SyncClient], use_local_files
...
+        setattr(client, "_log", original_log)
 
         # Create a closure to capture sync_client and use_local_files
-        def log_wrapper(self: Any, **kwargs) -> LogResponseType:
+        def log_wrapper(self: T, **kwargs) -> LogResponseType:
             return _overload_log(self, sync_client, use_local_files, **kwargs)
 
-        client.log = types.MethodType(log_wrapper, client)
+        # Replace the log method
+        setattr(client, "log", types.MethodType(log_wrapper, client))
```
peadaroh: Same q as above for these too

src/humanloop/overload.py
```diff
@@ -55,13 +59,13 @@ def _get_file_type_from_client(
...
-    raise ValueError(f"Unsupported client type: {type(client)}")
+    else:
+        raise ValueError(f"Unsupported client type: {type(client)}")
 
 
-def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, Any]:
+def _handle_tracing_context(kwargs: Dict[str, Any], client: T) -> Dict[str, Any]:
     """Handle tracing context for both log and call methods."""
-    trace_id = get_trace_id()
+    trace_id = get_trace_id()
```
harry-humanloop, line 68:
```suggestion
    trace_id = get_trace_id()
```

src/humanloop/overload.py
harry-humanloop: yeah this expansion into a getattr+setattr makes it harder to read - would push against this change (as it's not the type checker pushing us to better code)

AlePouroullis: Fair enough. Have addressed

harry-humanloop (approved):

PR #72, opened 2025-05-16
Release 0.8.40b5

andreibratu:
* Deprecate log_status on on Logs other than flow and agent log actions
* Updated dosctrings of Python functions
* Fixed regression in typing from sync overload


AlePouroullis:
Should be `0.8.40b5`

AlePouroullis (approved):

PR #73, opened 2025-05-16
Fix/consistent path normalization

AlePouroullis:
Most of the diffs are from rewriting/adding/removing tests. 

The core logic that changed was the handling of paths in the `.call()` and `.log()` operations when using local files.


src/humanloop/cli/__main__.py
```diff
@@ -81,6 +81,7 @@ def get_client(
     """
     if not api_key:
         api_key = load_api_key(env_file)
+    print(api_key)
```
peadaroh: Is this intentional?

src/humanloop/cli/__main__.py
```diff
@@ -218,6 +220,7 @@ def pull(
 
     Currently only supports syncing Prompt and Agent files. Other file types will be skipped."""
     client = get_client(api_key, env_file, base_url)
+    # Although pull() is available on the Humanloop client, we instantiate SyncClient separately as we need to control its log level
```
peadaroh: Could you elaborate on why we need to control log level? 

Does this have any implication on other things that happen in constructor related to OTel?

src/humanloop/client.py
```diff
@@ -431,6 +431,11 @@ def pull(self, path: Optional[str] = None, environment: Optional[str] = None) ->
 
         If you specify `local_files_directory="data/humanloop"`, files will be saved in ./data/humanloop/ instead.
 
+        Important note about paths:
+        - Paths should not contain leading or trailing slashes
+        - When specifying a file path, include the file extension (`.prompt` or `.agent`)
+        - When specifying a directory path, do not include a trailing slash
```
peadaroh: > When specifying a directory path, do not include a trailing slash

What happens if they do? Thought this would be something we should be able to santize

src/humanloop/cli/__main__.py
AlePouroullis: I had that there for debugging (my environment variables didn't seem to update after reloading window).
Good catch.

src/humanloop/overload.py
```diff
@@ -87,45 +89,65 @@ def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, An
...
+            f"Paths must follow the standard API format 'path/to/resource' without leading or trailing slashes. "
+            f"Please use '{normalized_path}' instead."
+        )
+
+    # Then check for file extensions
+    if sync_client.is_file(path):
+        try:
+            parts = path.rsplit(".", 1)
+            path_without_extension = parts[0] if len(parts) > 0 else path
+        except Exception:
```
peadaroh: Can you not be more specific on what type of exception you are defending against here?

src/humanloop/overload.py
```diff
@@ -173,7 +195,11 @@ def _overload_log(self: Any, sync_client: Optional[SyncClient], use_local_files:
 def _overload_call(self: Any, sync_client: Optional[SyncClient], use_local_files: bool, **kwargs) -> CallResponseType:
     try:
         kwargs = _handle_tracing_context(kwargs, self)
-        kwargs = _handle_local_files(kwargs, self, sync_client, use_local_files)
+        if use_local_files and _get_file_type_from_client(self) in SyncClient.SERIALIZABLE_FILE_TYPES:
+            if sync_client is None:
+                logger.error("sync_client is None but client has call method and use_local_files=%s", use_local_files)
```
peadaroh: This error message is hard to parse

src/humanloop/client.py
AlePouroullis: I realize this is incorrect here. `path` in the `pull()` operation is more liberal in terms of format—it normalizes/cleans up silently. `path` in `call()` and `log()` are strict, performing validation but no silent sanitization.

src/humanloop/path_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def normalize_path(path: str, strip_extension: bool = False) -> str:
+    """Normalize a path to the standard API format: path/to/resource, where resource can be a file or a directory."""
+    # Remove leading/trailing slashes
+    path_obj = Path(path.strip("/"))
+
+    # Check if path is absolute
+    if path_obj.is_absolute():
+        raise HumanloopRuntimeError("Absolute paths are not supported. Ensure the ")
```
peadaroh: Ensure the what?

src/humanloop/sync/sync_client.py
```diff
@@ -336,40 +330,53 @@ def pull(self, path: Optional[str] = None, environment: Optional[str] = None) ->
...
+            path = path.strip()
+            # Check if path has leading/trailing slashes
+            if path != path.strip("/"):
+                raise HumanloopRuntimeError(
+                    f"Invalid path: {path}. Path should not contain leading/trailing slashes. "
+                    f'Valid examples: "path/to/file.prompt" or "path/to/directory"'
+                )
+
+            # Check if it's a file path (has extension)
+            is_file_path = self.is_file(path)
```
PeadarOhAodha, line 347: is the variable 'is_file_path' used for anything subsequently here?

tests/custom/integration/test_local_file_operations.py
```diff
@@ -0,0 +1,314 @@
...
+        elif test_file.type == "agent":
+            humanloop_client.agents.call(
+                path=test_file.path,
+                version_id=test_file.version_id,
+                environment="staging",
+                messages=[{"role": "user", "content": "Testing"}],
+            )
+
+
+# def test_agent_local_file_usage(
```
peadaroh, line 260: Why is this commented out?

tests/custom/integration/test_sync.py
```diff
@@ -47,165 +32,71 @@ def test_pull_basic(
...
-
-
-def test_overload_version_environment_handling(
+    """Test error handling when environment doesn't exist."""
+    humanloop_client = get_humanloop_client(local_files_directory=str(tmp_path))
+    with pytest.raises(HumanloopRuntimeError, match="Environment .* does not exist"):
+        humanloop_client.pull(environment="invalid_environment")
+
+
+# def test_pull_with_environment(
```
peadaroh: why commented out?

src/humanloop/path_utils.py
AlePouroullis: I was meaning to modify this function to accept a base_dir that could be used to inform the user that their path should be relative to it. However, this only makes sense in the context of the sync client, not a generic util.

On review, I realize that this check is not necessary, and in fact erroneous: shouldn't be doing validation in a normalization function. And redundant, since we strip the leading/trailing slashes prior.

src/humanloop/overload.py
```diff
@@ -87,45 +89,65 @@ def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, An
...
     client: Any,
-    sync_client: Optional[SyncClient],
-    use_local_files: bool,
+    sync_client: SyncClient,
 ) -> Dict[str, Any]:
-    """Handle local file loading if enabled."""
-    if not use_local_files or "path" not in kwargs or sync_client is None:
-        return kwargs
-
+    """Handle local file loading."""
```
harry-humanloop: docstring needs elaboration - what does "handle" mean? what's the returned dict? what's "kwargs"?

src/humanloop/overload.py
```diff
@@ -87,45 +89,65 @@ def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, An
...
     if "id" in kwargs:
         raise HumanloopRuntimeError("Can only specify one of `id` or `path`")
 
+    path = kwargs["path"]
+
+    # First check for path format issues (absolute paths or leading/trailing slashes)
+    normalized_path = path.strip("/")
```
harry-humanloop, line 121: I'm surprised this is necessary. why are we doing something so manually?
also: https://docs.python.org/3/library/os.html#os.sep

src/humanloop/overload.py
```diff
@@ -87,45 +89,65 @@ def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, An
...
+        raise HumanloopRuntimeError(
+            f"Path '{path}' format is invalid. "
+            f"Paths must follow the standard API format 'path/to/resource' without leading or trailing slashes. "
+            f"Please use '{normalized_path}' instead."
+        )
+
+    # Then check for file extensions
+    if sync_client.is_file(path):
+        try:
+            parts = path.rsplit(".", 1)
```
harry-humanloop: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

should avoid doing this path string processing imo

src/humanloop/overload.py
harry-humanloop: +1 for avoiding bare exceptions - and this should've had a comment.

(think this can be avoided with path.stem as linked above)

src/humanloop/overload.py
```diff
@@ -87,45 +89,65 @@ def _handle_tracing_context(kwargs: Dict[str, Any], client: Any) -> Dict[str, An
...
+            path_without_extension = parts[0] if len(parts) > 0 else path
+        except Exception:
+            path_without_extension = path
+
+        raise HumanloopRuntimeError(
```
harry-humanloop, line 135: am i misreading the indentation here?
this looks like it always raises an error if `sync_client.is_file(path)`

src/humanloop/overload.py
```diff
@@ -151,11 +173,11 @@ def _overload_log(self: Any, sync_client: Optional[SyncClient], use_local_files:
         kwargs = _handle_tracing_context(kwargs, self)
 
         # Handle local files for Prompts and Agents clients
-        if _get_file_type_from_client(self) in ["prompt", "agent"]:
+        if use_local_files and _get_file_type_from_client(self) in SyncClient.SERIALIZABLE_FILE_TYPES:
             if sync_client is None:
                 logger.error("sync_client is None but client has log method and use_local_files=%s", use_local_files)
```
harry-humanloop: what does "client has log method" mean?

src/humanloop/path_utils.py
```diff
@@ -0,0 +1,16 @@
...
+    path_obj = Path(path.strip("/"))
+
+    # Handle extension
+    if strip_extension:
+        normalized = str(path_obj.with_suffix(""))
+    else:
+        normalized = str(path_obj)
+
+    # Normalize separators
+    return "/".join(part for part in normalized.replace("\\", "/").split("/") if part)
```
harry-humanloop: is this https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

src/humanloop/path_utils.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+
+
+def normalize_path(path: str, strip_extension: bool = False) -> str:
+    """Normalize a path to the standard API format: path/to/resource, where resource can be a file or a directory."""
```
harry-humanloop: should be expanded with what this is for, and a few examples imo.

i can't tell why we strip leading/trailing slashes and whether that's correct.

tests/custom/unit/test_path_utils.py
```diff
@@ -0,0 +1,28 @@
+from humanloop import path_utils
+
+
+def test_normalize_path():
+    """Test path normalization functionality."""
+    # GIVEN various file paths with different formats
+    test_cases = [
```
harry-humanloop: should this use pytest.mark.parametrize? https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize-basics

would give the three items in the tuples better variable names

PR #75, opened 2025-05-21
Introduce async decorators

andreibratu:
* File decorators now have async equivalents: a_flow, a_tool and a_prompt
* They can be used in evals.run().
* We don't have a strong signal async decorators yet; thus the implementation places them on the sync client, leverages existing custom code for `evals.run()`, and makes 1-3 blocking calls per decorated function call.
Implementing a fully non-blocking version of the decorators implies rewriting the overload, OTEL exporter and eval run logic (from threaded to non-blocking), requiring a lot of effort.
* I propose scoping out the full reimplementation of custom functionality: The biggest non-blocking gains come from supporting async user code.
* OTEL offers the same API for spans, so I assume the context and span creation operations are in-memory and cheap

### DIFF
--- a/src/humanloop/overload.py
+++ b/src/humanloop/overload.py
@@ -264,7 +264,7 @@ def overload_client(
         def log_wrapper(self: T, **kwargs) -> LogResponseType:
             return _overload_log(self, file_syncer, use_local_files, **kwargs)
 
-        client.log = types.MethodType(log_wrapper, client)  # type: ignore [method-assign, union-attr, assignment]
+        client.log = types.MethodType(log_wrapper, client)  # type: ignore [method-assign]
 
     # Overload call method for Prompt and Agent clients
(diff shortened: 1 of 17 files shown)


src/humanloop/overload.py
```diff
@@ -264,7 +264,7 @@ def overload_client(
         def log_wrapper(self: T, **kwargs) -> LogResponseType:
             return _overload_log(self, file_syncer, use_local_files, **kwargs)
 
-        client.log = types.MethodType(log_wrapper, client)  # type: ignore [method-assign, union-attr]
+        client.log = types.MethodType(log_wrapper, client)  # type: ignore [method-assign, union-attr, assignment]
```
AlePouroullis: Only `method-assign` needs to be ignored

AlePouroullis (approved):

src/humanloop/overload.py
andreibratu: ack

